PR #1, opened 2022-03-11
Update build script

williamputraintan:
- Added/Modify build layer script (use `make build` to build all layers)
- Added `make install`
- Change from `aws-sam-cli-beta-cdk` to `aws-sam-cli`. As now supported in [SAM cli](https://docs.aws.amazon.com/cdk/v2/guide/sam.html)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ install:
 	npm install -g aws-cdk@2.15.0
 	
 	brew tap aws/tap
-	brew install aws-sam-cli@1.40.1
+	brew reinstall aws-sam-cli
 build:
 #   Build lambda Layer
 	for dir in $(shell find ./lambdas/layers -maxdepth 1 -mindepth 1 -type d -exec basename {} \;); do ./lambdas/layers/create_layer_package.sh $$dir; done
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,9 @@
+install:
+	pip install -r requirements.txt       # pulls in CDKv2 `aws-cdk-lib` and `construct` libs
+	npm install -g aws-cdk@2.15.0
+	
+	brew tap aws/tap
+	brew install aws-sam-cli@1.40.1
 build:
 #   Build lambda Layer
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,6 +1,16 @@
...
 build:
-	sam-beta-cdk build
+#   Build lambda Layer
+	for dir in $(shell find ./lambdas/layers -maxdepth 1 -mindepth 1 -type d -exec basename {} \;); do ./lambdas/layers/create_layer_package.sh $$dir; done
+#   Build Cdk to cdk.out
+	cdk synth
 deploy: build
-	cdk deploy -a .aws-sam/build --profile ${AWS_PROFILE}
+	cdk deploy UmccrEventBus --profile ${AWS_PROFILE}
+	cdk deploy UmccrEventBusSchemaStack --profile ${AWS_PROFILE}
```
reisingerf, line 14: What's the reason for `--profile ${AWS_PROFILE}`?

Do we need/want to restrict credential resolution options?

Makefile
```diff
@@ -1,6 +1,16 @@
+install:
+	pip install -r requirements.txt       # pulls in CDKv2 `aws-cdk-lib` and `construct` libs
+	npm install -g aws-cdk@2.15.0
```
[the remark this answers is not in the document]
williamputraintan: Was added to make sure that the current version should not keep on changing and reaches to the stage if the version is no longer backward-compatible with the code. But the version could be removed.

Makefile
williamputraintan: Yes could removed. Was thinking of allowing to deploy on a certain profile (allowing to choose for `prod` or `dev`) when necessary. But I guess could be solve when by default setting it on environment variable. 

Good to remove, specially when building on codebuild it might not have `AWS_PROFILE` as their environment variable

victorskl (approved):
Okay! Let push it through for now. We will probably go rounds of iteration on this. Of course.

PR #10, opened 2022-04-05
Improve project setup

victorskl:
* Rebranding to OrcaBus
* Replaced `setup.py` with `requirements.txt`
* Bumped Python 3.9


reisingerf (approved):

.pre-commit-config.yaml
```diff
@@ -1,21 +1,24 @@
...
-      - id: detect-aws-credentials
+#      - id: detect-aws-credentials
       - id: detect-private-key
       - id: end-of-file-fixer
       - id: trailing-whitespace
-      - id: no-commit-to-branch
-        args: [ --branch, main]
+#      - id: no-commit-to-branch
+#        args: [ --branch, main, --branch, master, --pattern, release/.* ]
+
```
reisingerf, line 15: Just curious: any particular reason why you took those out?

victorskl: Ah DevOops! Fixing

williamputraintan (approved):

.pre-commit-config.yaml
reisingerf: And the aws creds check?

victorskl: Yep; that `id: detect-aws-credentials` hook won't work for SSO-ed CDK. Reproducible: try `rm ~/.aws/credentials` and `make check`. Hence, Yelp `detect-secrets` which cover that plus more heuristic scan vectors.

victorskl: i.e. `cd orcabus && detect-secrets scan --list-all-plugins`

victorskl: Actually, my memory comes back now; why I commented `no-commit-to-branch` hook initially. For this, to work, we either 

A. Better should have `dev` branch as default upstream. 
B. Otherwise that hook scanning hit to the local `main` & block (which is not intended)

Perhaps, we better just wish protect direct push to remote i.e. better be, if we rather just want to stick with a `single` branch strategy.

```
      - id: no-commit-to-branch
        args: [ --branch, origin/main, --branch, origin/master, --pattern, origin/release/.* ]
```

PR #13, opened 2023-04-18
Refactored project setup on initial structure

victorskl:
* Changed CDK to be coded in TypeScript
* Followed CDK CodePipeline best practise and reference architecture
  by introducing concepts: workload, stateful, stateless, component
* Split CDK entry points: manual and self-mutating pipeline
* Initial mark up for unit tests structure for both TS/Py
  * deploy time code in TypeScript
  * run time code in Python


victorskl:
Better to be viewed it at its tree - https://github.com/umccr/orcabus/tree/refactor-setup

victorskl:
[Orcabus V1: Restructure setup following CDK Pipeline Reference Architecture](https://trello.com/c/MWwjUOvQ/1418-orcabus-v1-restructure-setup-following-cdk-pipeline-reference-architecture)

lib/workload/stateless/layers/component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  private _build_deps() {
+    // TODO using docker SDK to build the deps as part of `cdk synth`, if some `build-auto` flag is true
+    //  See https://github.com/umccr/infrastructure/blob/2a1d47c485d11f8a4a9bf0d2cd865f8450164876/cdk/apps/htsget/htsget/goserver.py#L374
+  }
+
+  private createLambdaLayer(name: string) {
+    return new aws_lambda.LayerVersion(this, 'OrcaBus_' + name + '_LayerVersion', {
+      code: aws_lambda.Code.fromAsset(path.join(__dirname, name + '.zip')),
+      compatibleRuntimes: [aws_lambda.Runtime.PYTHON_3_9],
```
andrewpatto, line 27: 3.10? (given the other references through out to 3.10)

cdk.json
```diff
@@ -1,39 +1,51 @@
...
+    "@aws-cdk/aws-route53-patters:useCertificate": true,
+    "@aws-cdk/customresources:installLatestAwsSdkDefault": false,
+    "@aws-cdk/aws-rds:databaseProxyUniqueResourceName": true,
+    "@aws-cdk/aws-codedeploy:removeAlarmsFromDeploymentGroup": true,
+    "@aws-cdk/aws-apigateway:authorizerChangeDeploymentLogicalId": true,
+    "@aws-cdk/aws-ec2:launchTemplateDefaultUserData": true,
+    "@aws-cdk/aws-secretsmanager:useAttachedSecretResourcePolicyForSecretTargetAttachments": true,
+    "@aws-cdk/aws-redshift:columnId": true,
+    "@aws-cdk/aws-stepfunctions-tasks:enableEmrServicePolicyV2": true
   }
```
andrewpatto, line 50: Is there any of these where we care what the result is? Wouldn't it be better to leave it at its defaults unless we explicilty need something different?

README.md
```diff
@@ -2,95 +2,75 @@
...
-```
-
-### Pre-commit hooks
-
-We are using pre-commit hooks to check for any secret-key leaks and for linting purposes.
-Check [UMCCR wiki](https://github.com/umccr/wiki/blob/master/computing/dev-environment/git-hooks.md) for details.
-
-```
-brew install pre-commit detect-secrets
+conda create -n orcabus python=3.10
```
andrewpatto, line 13: what's condas role is all this?

lib/workload/stateless/layers/component.ts
victorskl: Ah good catch! I will fix it.

cdk.json
victorskl: Eh, another good point. Thanks. TBH, I wasn't sure. 😬 These entries were populated as part of `cdk init`. 

Sure, I heed your advice and, I will remove them, I reckon; I'll wait bit & lazy consensus.

README.md
victorskl: Conda is mainly for local dev Python environment. It could be `python3 -m venv venv` alternatively; sure.

andrewpatto: Doesn't it need to match whatever package manager is being used for the individual python lambdas? If people are installing things into a generic conda env their actual deployed code may not have the same packages?

victorskl: Yes, it needs to match. It is more like; a high level contract/assumption such that in the current product iteration life cycle, the major version must any of 3.10 tree.

Package dependencies versions are pinned. They must work in some generic Python 3.10 environment as well as in the target Lambda runtime; which should be 3.10 as well. While having mindful with any native binding Python package in use.

CodePipeline unit/integration test cases to reinforce this.

cdk.json
victorskl: Cleared away CDK context with https://github.com/umccr/orcabus/commit/d1ed4fb9671cc6aff896fa99c3cbb39db565da1a

lib/workload/stateless/layers/component.ts
victorskl: Externalised the Lambda runtime Python version config with https://github.com/umccr/orcabus/commit/2f70421ca7df0051b32f35d0394b4e43b52a486e

reisingerf (approved):

lib/workload/stateful/database/component.ts
```diff
@@ -0,0 +1,47 @@
...
+      }),
+      parameterGroup: aws_rds.ParameterGroup.fromParameterGroupName(
+        this,
+        id + 'ParameterGroup',
+        props.parameterGroupName
+      ),
+      removalPolicy: RemovalPolicy.DESTROY,
+      scaling: {
+        // autoPause: Duration.seconds(60),
+        minCapacity: aws_rds.AuroraCapacityUnit.ACU_1,
```
mmalenic, line 39: Is this for serverless v1 or v2? I think v2 is not fully supported in CDK yet.

You could potentially use serverless v2 by manually specifying the `aws_rds.CfnDBCluster` engineMode and creating a writer instance, similar to [here](https://github.com/umccr/elsa-data-aws-infrastructure/blob/dc19291241e73618c350665b0b0340fb55ff987d/rds/serverless-base-database.ts#L60-L71) and [here](https://github.com/umccr/elsa-data-aws-infrastructure/blob/dc19291241e73618c350665b0b0340fb55ff987d/rds/serverless-base-database.ts#L88-L94). That way, you could set the minCapacity to 0.5, which might be better? Although, it would scale slower to a higher ACU then.

mmalenic (approved):

williamputraintan (approved):

lib/workload/stateful/database/component.ts
victorskl: Thanks for pointers, Marko. Absolutely we gotta do that. Here is, just a placeholder code for now, yes.

victorskl:
Thanks all! Merging...

PR #19, opened 2023-05-04
Implemented externalise config constants pass-down as props pattern

victorskl:
* Minor refactored and improved arrangement

### DIFF
--- a/config/constants.ts
+++ b/config/constants.ts
@@ -1,0 +1,48 @@
+import { OrcaBusStatefulConfig } from '../lib/workload/orcabus-stateful-stack';
+import { AuroraPostgresEngineVersion } from 'aws-cdk-lib/aws-rds';
+import { OrcaBusStatelessConfig } from '../lib/workload/orcabus-stateless-stack';
+
+const regName: string = 'OrcaBusSchemaRegistry';
+
+export const orcaBusStatefulConfig: OrcaBusStatefulConfig = {
+  schemaRegistryProps: {
+    registryName: regName,
+    description: 'Registry for OrcaBus Events',
+  },
+  eventBusProps: {
+    eventBusName: 'OrcaBusMain',
+    archiveName: 'OrcaBusMainArchive',
+    archiveDescription: 'OrcaBus main event bus archive',
+    archiveRetention: 365,
+  },
+  databaseProps: {
+    clusterIdentifier: 'orcabus-db',
+    defaultDatabaseName: 'orcabus',
+    version: AuroraPostgresEngineVersion.VER_14_6,
+    parameterGroupName: 'default.aurora-postgresql14',
+    username: 'admin',
+  },
+  securityGroupProps: {
+    securityGroupName: 'OrcaBusLambdaSecurityGroup',
+    securityGroupDescription: 'Allow within same SecurityGroup',
+  },
+};
+
+export const orcaBusStatelessConfig: OrcaBusStatelessConfig = {
+  multiSchemaConstructProps: {
+    registryName: regName,
+    schemas: [
+      {
+        schemaName: 'BclConvertWorkflowRequest',
+        schemaDescription: 'Request event for BclConvertWorkflow',
+        schemaLocation: __dirname + '/event_schemas/BclConvertWorkflowRequest.json',
+      },
+      {
+        schemaName: 'DragenWgsQcWorkflowRequest',
+        schemaDescription: 'Request event for DragenWgsQcWorkflowRequest',
+        schemaLocation: __dirname + '/event_schemas/DragenWgsQcWorkflowRequest.json',
+      },
+    ],
+  },
+  bclConvertFunctionName: 'orcabus_bcl_convert',
+};
--- a/config/constants.ts
+++ b/config/constants.ts
@@ -3,6 +3,8 @@ import { AuroraPostgresEngineVersion } from 'aws-cdk-lib/aws-rds';
 import { OrcaBusStatelessConfig } from '../lib/workload/orcabus-stateless-stack';
 
 const regName: string = 'OrcaBusSchemaRegistry';
+const eventBusName: string = 'OrcaBusMain';
+const lambdaSecurityGroupName: string = 'OrcaBusLambdaSecurityGroup';
 
 export const orcaBusStatefulConfig: OrcaBusStatefulConfig = {
   schemaRegistryProps: {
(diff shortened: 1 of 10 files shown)


victorskl:
[Orcabus V1: Implement externalise config for CDK constructs](https://trello.com/c/4hmvdUDX/1422-orcabus-v1-implement-externalise-config-for-cdk-constructs)

config/constants.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const regName: string = 'OrcaBusSchemaRegistry';
+
+export const orcaBusStatefulConfig: OrcaBusStatefulConfig = {
+  schemaRegistryProps: {
+    registryName: regName,
+    description: 'Registry for OrcaBus Events',
+  },
+  eventBusProps: {
+    eventBusName: 'OrcaBusMain',
```
reisingerf, line 13: I guess that will be used at some point in the stateless config, so probably good to pull out into global const?

config/constants.ts
```diff
@@ -0,0 +1,48 @@
...
+    archiveName: 'OrcaBusMainArchive',
+    archiveDescription: 'OrcaBus main event bus archive',
+    archiveRetention: 365,
+  },
+  databaseProps: {
+    clusterIdentifier: 'orcabus-db',
+    defaultDatabaseName: 'orcabus',
```
reisingerf, line 20: Same as for the bus. I suppose the DB name will be needed as well in the stateless stack?

config/constants.ts
```diff
@@ -0,0 +1,48 @@
...
+    version: AuroraPostgresEngineVersion.VER_14_6,
+    parameterGroupName: 'default.aurora-postgresql14',
+    username: 'admin',
+  },
+  securityGroupProps: {
+    securityGroupName: 'OrcaBusLambdaSecurityGroup',
```
reisingerf, line 26: Needed in stateless?

lib/workload/orcabus-stateless-stack.ts
```diff
@@ -3,50 +3,55 @@ import { aws_ec2 } from 'aws-cdk-lib';
...
-    //     this,
-    //     'BclConvertLambdaSg',
-    //     SecurityGroupConstruct.ORCABUS_LAMBDA_SECURITY_GROUP,  // FIXME externalise config
-    //     vpc,
-    //   ),
-    // ];
+    const securityGroups = [
+      aws_ec2.SecurityGroup.fromLookupByName(
+        this,
+        'LambdaSecurityGroup',
```
reisingerf, line 28: Is this supposed to be `OrcaBusLambdaSecurityGroup` from the stateful stack?

lib/workload/stateful/securitygroup/component.ts
```diff
@@ -2,21 +2,21 @@ import { Construct } from 'constructs';
...
-  public static readonly ORCABUS_LAMBDA_SECURITY_GROUP: string = 'OrcaBusLambdaSecurityGroup';
 
-  constructor(scope: Construct, id: string, props: Props) {
+  constructor(scope: Construct, id: string, vpc: IVpc, props: SecurityGroupProps) {
     super(scope, id);
 
-    const lambdaSecurityGroup = new SecurityGroup(this, 'LambdaSecurityGroup', {
-      securityGroupName: SecurityGroupConstruct.ORCABUS_LAMBDA_SECURITY_GROUP,
-      vpc: props.vpc,
+    const lambdaSecurityGroup = new SecurityGroup(this, id + 'LambdaSecurityGroup', {
```
reisingerf, line 15: Ditto `OrcaBusLambdaSecurityGroup`?

lib/workload/stateless/schema/component.ts
```diff
@@ -20,23 +11,23 @@ export interface SchemaConstructProps {
...
 
-    const content: string = this.getSchemaContent(cProps.schemaLocation)
+    const content: string = this.getSchemaContent(props.schemaLocation);
 
-    new eventschemas.CfnSchema(this, cProps.schemaName, {
+    new eventschemas.CfnSchema(this, props.schemaName, {
       content: content,
-      registryName: cProps.registryName,
+      registryName: props.registryName,
       type: 'OpenApi3',
```
reisingerf, line 26: Could push this into props as well. Make it optional and default to this value (possible use case for optional config + defaults)

reisingerf (commented):
Nice!

lib/workload/stateful/securitygroup/component.ts
victorskl: Yup. Same explanation. The second argument is for Construct ID. 

Referencing to stateful security group name is happening in next line, as part of `props`.

lib/workload/stateless/schema/component.ts
victorskl: You mean `type: 'OpenApi3'`? Sure, happy to update that right away, yes.

config/constants.ts
reisingerf: Hm... good question... 
I thought about high level constant, along the same line as
```const regName: string = 'OrcaBusSchemaRegistry';```

It is true that the value is available from the stateful stack config....
I thought with a separate high level constant the stateless stack could be independent of the stateful stack or its props structure, but the two usually go hand in hand so a bit of dependency is expected/acceptable I guess. 
Can always change down the road....

lib/workload/orcabus-stateless-stack.ts
reisingerf: Ah, yes!
Read too quickly...

lib/workload/stateless/schema/component.ts
reisingerf: Well, since the schema are stateless and defined outside, so this should probably be configurable. 
Or would we want to enforce a particular type?

config/constants.ts
victorskl: Ah, like `regName` var. Right. I know what you mean. Ok, I will update that to realign them.

lib/workload/stateless/schema/component.ts
victorskl: Ok no prob; I will update it to externalise it as well.

config/constants.ts
victorskl: Good point! It might be more like db SM that need externalise.. I will investigate

lib/workload/stateless/schema/component.ts
victorskl: Added with https://github.com/umccr/orcabus/commit/af853c6097cab635d74df1a23a97bffead6ce212

reisingerf (approved):

victorskl:
Merging...

PR #22, opened 2023-05-23
Serverless MySQL

williamputraintan:
- Attempt on changing to Serverless MySQL

### DIFF
--- a/bin/orcabus.ts
+++ b/bin/orcabus.ts
@@ -14,6 +14,12 @@ const props: cdk.StackProps = {
   },
 };
 
-new OrcaBusStatefulStack(app, 'OrcaBusStatefulStack', { ...props, ...orcaBusStatefulConfig });
+const orcaBusStatefulStack = new OrcaBusStatefulStack(app, 'OrcaBusStatefulStack', {
+  ...props,
+  ...orcaBusStatefulConfig,
+});
 
-new OrcaBusStatelessStack(app, 'OrcaBusStatelessStack', { ...props, ...orcaBusStatelessConfig });
+const orcaBusStatelessStack = new OrcaBusStatelessStack(app, 'OrcaBusStatelessStack', {
+  ...props,
+  ...orcaBusStatelessConfig,
+});
--- a/bin/orcabus.ts
+++ b/bin/orcabus.ts
@@ -14,12 +14,12 @@ const props: cdk.StackProps = {
   },
 };
 
-const orcaBusStatefulStack = new OrcaBusStatefulStack(app, 'OrcaBusStatefulStack', {
+new OrcaBusStatefulStack(app, 'OrcaBusStatefulStack', {
   ...props,
   ...orcaBusStatefulConfig,
 });
 
-const orcaBusStatelessStack = new OrcaBusStatelessStack(app, 'OrcaBusStatelessStack', {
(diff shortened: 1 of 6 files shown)


bin/orcabus.ts
```diff
@@ -14,6 +14,12 @@ const props: cdk.StackProps = {
   },
 };
 
-new OrcaBusStatefulStack(app, 'OrcaBusStatefulStack', { ...props, ...orcaBusStatefulConfig });
+const orcaBusStatefulStack = new OrcaBusStatefulStack(app, 'OrcaBusStatefulStack', {
+  ...props,
+  ...orcaBusStatefulConfig,
+});
 
-new OrcaBusStatelessStack(app, 'OrcaBusStatelessStack', { ...props, ...orcaBusStatelessConfig });
```
williamputraintan, line 19: Before changing more on this, instead of passing props of string (e.g. eventBus name, SG name) I think we should just pass the resource itself, so it doesn't need for a lookup anymore. (As it is on the example of [Stateless vs Stateful](https://blog.serverlessadvocate.com/serverless-aws-cdk-pipeline-best-practices-patterns-part-1-ab80962f109d) blog)

Or was this is done on purpose ?

lib/workload/stateful/securitygroup/component.ts
```diff
@@ -1,25 +1,33 @@
...
   securityGroupDescription: string;
 }
-
+export interface SecurityGroupConstructProps extends SecurityGroupProps {
+  dbSecurityGroup: ec2.SecurityGroup;
+}
 export class SecurityGroupConstruct extends Construct {
-  constructor(scope: Construct, id: string, vpc: IVpc, props: SecurityGroupProps) {
+  constructor(scope: Construct, id: string, vpc: ec2.IVpc, props: SecurityGroupConstructProps) {
     super(scope, id);
```
williamputraintan, line 14: Just wondering here, if the SG is purposely for lambda, would it be better if this is created right on the stack where lambda will be created (on the `OrcaBusStatelessStack` before the workflow construct)?

williamputraintan:
[Orcabus V1: Implement RDS Aurora Serverless v2 database](https://trello.com/c/BYN0Voe6/1435-orcabus-v1-implement-rds-aurora-serverless-v2-database)

bin/orcabus.ts
victorskl: Ah. That blog post. 😅 It is on purpose, yes, Will. Because we do not want to pass around "resource reference". Considering situation when some consuming Construct get destroy (i.e. Stateless/self-mutating part), we do not want to either get deleted stateful resource by accident nor get entangled not able to destroy Stateless Construct. So, falling this back to the lowest primitive possible as config string or number.

lib/workload/stateful/securitygroup/component.ts
victorskl: You are right, there will be another App specific SG down there in Stateless Construct for other inbound/outbound rules need. 

However. This SG is special for Lambda to Database with VPC. Because the DB will be placed in Isolated subnet i.e. air-gapped. No direct access from Internet nor external but must be within VPC only.

victorskl:
Thanks Will for PR. I will review soon, tomorrow arvo, probably.

victorskl (approved):
Thanks Will. LGTM! Please squash merge.

reisingerf (approved):
LGTM

PR #23, opened 2023-05-24
Refactored OrcaBus Stateless component with MicroService

victorskl:
* Between microservice app, share nothing or share as little as possible
* Each microservice belong to one bounded context and responsible for that domain
* Promote rapid prototype dev tooling with Skel generator for boilerplate
* Added Django based Skel Profile and CDK Construct Library for rapid starter
* Added docker compose stack with localstack for AWS and ICA mock services
* Added documentation and improved project setup


reisingerf (approved):
Wow, that will take me a while to get through, but looks great!

PR #27, opened 2023-06-09
FileManager microservice Rust skel

brainstorm:
... and initial draft/incomplete definition of filemanager microservice client API.


victorskl (commented):
Nice. LGTM so far; will review in more once PR is finalised! 👍

skel/rust-api/src/filemanager.rs
```diff
@@ -0,0 +1,51 @@
...
+use serde::{Deserialize, Serialize};
+use utoipa::{IntoParams, ToSchema};
+
+use tracing::{ info };
+//type Store = Mutex<Vec<FileManager>>;
+// TODO: SQLx is the store backend, through db.rs
+
+/// Item to do.
+#[derive(Serialize, Deserialize, ToSchema, Clone)]
+pub struct FileManager {
```
reisingerf, line 11: Is that the Rust equivalent of a Django Model class? 
Just wondering about the naming... the FileManager is more the microservice itself where the model would be more something like File/Object/FileObject/... ?

Not at all into Rust, so please ignore it that does not apply!

reisingerf:
Great start, thanks!
(expect stupid noob Rust questions though ;-) )

skel/rust-api/src/filemanager.rs
brainstorm: Yes! I was torn between "File" and "Object", so I left it as "FileManager", which is clearly confusing and wrong, which one would you pick from the former two names? I'll change it :)

brainstorm: No `File` doesn't clash with names, Rust is pretty good with shadowing names without clashes... let's go with file then.

victorskl:
[OrcaBus V1: Design and Implement File Manager](https://trello.com/c/OmhmlAkN/1426-orcabus-v1-design-and-implement-file-manager)

victorskl (approved):
Rust skel looks good and LGTM! Happy to merge this ahead first as we would need to base OpenAPI doc arrangement from this PR.

skel/rust-api/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+            (name = "File", description = "File manager API")
+        )
+    )]
+    struct ApiDoc;
+
+    #[derive(Clone)]
+    struct AppState {}
+    struct SecurityAddon;
+
+    impl Modify for SecurityAddon {
```
mmalenic, line 48: Would it be better if these were outside of the main function?

brainstorm: Thought exactly the same right after merging it, along with putting a few more comments to guide new user's edits :-S

reisingerf (commented):
LATM

skel/rust-api/src/main.rs
brainstorm: https://github.com/umccr/orcabus/commit/483226ff149feddfd155e82b3a9823ef60df1752

PR #28, opened 2023-06-14
Implemented Sequence Run Manager microservice app

victorskl:
* The initial model based on Portal Sequence model with additional
  improvement made such as stronger mandatory fields and the only
  unique for field `instrument_run_id`
* Renamed fields so that it can be used for general Sequencing tech
  or, some plain raw data from S3 bucket location
* Added SRM.md developer documentation


victorskl:
[OrcaBus V1: Implement SequenceRunManager microservice app](https://trello.com/c/TQX8yJuc/1445-orcabus-v1-implement-sequencerunmanager-microservice-app)

lib/workload/stateless/sequence_run_manager/src/sequence_run_manager_proc/lambdas/bssh_event.py
```diff
@@ -0,0 +1,81 @@
+import os
+
+import django
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "sequence_run_manager.settings.base")
```
reisingerf, line 5: Do we need to set a default? 
Seems dangerous considering that the base config states it should not be used in production?

reisingerf (approved):

lib/workload/stateless/sequence_run_manager/src/sequence_run_manager_proc/lambdas/bssh_event.py
victorskl: Base as default is fine, Flo. In our deployment situation.
Those comments were a bit legacy, but good to know as dev note.

victorskl:
Merging...

PR #31, opened 2023-06-16
Improved Django Skel by removing base setting as default

victorskl:
* Deployed target must set `DJANGO_SETTINGS_MODULE` env var
  to `aws` setting. Else, it is better that let it crash by default.
* See #28 convo.


victorskl:
[OrcaBus V1: Implement SequenceRunManager microservice app](https://trello.com/c/TQX8yJuc/1445-orcabus-v1-implement-sequencerunmanager-microservice-app)

reisingerf (approved):

PR #32, opened 2023-06-21
Improved GHA CI build and local dev DX

victorskl:
* Bumped localstack v2
* Added reusable GHA container image caches.yml workflow
* Added mocking section in README

PR #34, opened 2023-06-21
Improved GHA CI container image caching with composite action

victorskl:


victorskl:
This is temporary debug branch to refactor reusable workflow into composite action from the main PR #32

I will merge this into its parent PR #32 and we shall continue review/comments over there.

FYI

PR #35, opened 2023-06-21
Init Metadata Manager

williamputraintan:
An init migration of the Metadata Model
- Migrating metadata from data-portal-api to this manager
- Add timestamp for each record created, and allow to save all history changes at every sync to the Django model
- Some unit-tests to support the model


victorskl:
[OrcaBus V1: Implement Metadata Manager microservice application](https://trello.com/c/gkIJiF0k/1482-orcabus-v1-implement-metadata-manager-microservice-application)

victorskl:
Thanks Will; I will review in tomorrow soon.

Meanwhile, when PR #32 get merged into the main, please rebase it. 
Then, it should take care of container image pull rate throttle issue by using [GHA action cache](https://github.com/umccr/orcabus/actions/caches). 🤞

victorskl (approved):
Looking good for first cut. So, go merge it, if any. Squash merge if you can.

Please do check with @reisingerf whether this should now be called LibraryManager or otherwise...

Also. We can revisit/tinkering about whether keep doing truncate & reload or, change-data-capture CDC with differential data loading...

Just pull up note reminder. No blocker. Leave it to you for follow up! 😃

lib/workload/stateless/metadata_manager/src/metadata_manager/urls/base.py
```diff
@@ -0,0 +1,12 @@
+from django.urls import path, include
+
+from metadata_manager.routers import OptionalSlashDefaultRouter
+from metadata_manager.viewsets.metadata import MetadataViewSet
+
+router = OptionalSlashDefaultRouter()
+router.register(r"metadata", MetadataViewSet, basename="metadata")
+
+urlpatterns = [path("iam/", include(router.urls)), path("", include(router.urls))]
```
reisingerf, line 9: Do we need the "iam/" path?

lib/workload/stateless/metadata_manager/src/metadata_manager/wsgi.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+from django.core.wsgi import get_wsgi_application
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "metadata_manager.settings.base")
```
reisingerf, line 5: I think we removed the default Django setting assuming it's already set on the ENV.

lib/workload/stateless/metadata_manager/src/metadata_manager_proc/lambdas/metadata_proc.py
```diff
@@ -0,0 +1,59 @@
+import os
+import django
+
+import logging
+from datetime import datetime
+from django.utils import timezone
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "metadata_manager.settings.base")
```
reisingerf, line 8: Ditto: assume set on ENV already.

lib/workload/stateless/metadata_manager/src/migrate.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+"""migrate lambda module
+
+Convenience AWS lambda handler for Django database migration command hook
+"""
+import os
+
+from django.core.management import execute_from_command_line
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "metadata_manager.settings.base")
```
reisingerf, line 10: Ditto assume already set.

reisingerf (approved):
Nice work! A great start!

Minor things to follow up on:
We should rename this from metadata to library as discussed and mentioned by Victor. 
The Django config settings were also discussed with Victor and I think we decided to remove the defaults to avoid possible misconfiguration if the env is not explicitly set up.

lib/workload/stateless/metadata_manager/src/metadata_manager/urls/base.py
victorskl: Yup, can omit `iam/` for now. Will; could you also update in skel profiles laster, pls.

lib/workload/stateless/metadata_manager/src/metadata_manager/wsgi.py
victorskl: Yup, skel updates with #31

PR #37, opened 2023-06-27
Filemanager: S3 events persistence

brainstorm:
* Serializes S3 event attributes to DB backend.
* Using Axum (and AWS SDK) for local development, will transition to lambda Rust runtime crates later.


brainstorm:
@mmalenic I was about to merge this gigantic PR but thought about getting your thumbsup first.

mmalenic (approved):
Yes, go for it.

In relation to the log groups, yes it fails if no events have been created yet. Maybe this needs a clearer note in the README.

brainstorm:
> In relation to the log groups, yes it fails if no events have been created yet. Maybe this needs a clearer note in the README.

More than a note: I meant that if a newcomer to the codebase reads the README, it should have the least amount of gotchas. Following the steps in sequence should yield a deterministic result and have a locally deployed dev setup as a result.

Issue #70, opened 2023-12-11
filemanager: add database infrastructure 


mmalenic:
The deployment code for filemanager does not link to a database yet:
https://github.com/umccr/orcabus/blob/492c4c456695b97354212fdea97ab77e16d7d861/lib/workload/stateful/filemanager/deploy/stack/filemanager_stack.ts#L103
Any filemanager Lambda will expect database access to query and ingest file events, which for now is postgres. This functionality should be added to the CDK code.

mmalenic:
Closed via #74

Issue #71, opened 2023-12-11
filemanager: setup testing mocks


mmalenic:
Since we are removing localstack, we should include mock objects to do local tests before deploying to dev. Ideally, it should be possible to run `cargo test` to quickly check unit tests. The docker postgres database can be used as a local database connection, or sqlx already has [some testing](https://docs.rs/sqlx/latest/sqlx/attr.test.html) functionality. However, the AWS clients will need to be mocked to create unit tests, similar to how it's described [here](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/testing.html). Currently, this includes `aws_sdk_sqs::Client` and `aws_sdk_s3::Client`.

PR #74, opened 2023-12-13
Filemanager: fix ingestion/deployment

brainstorm:
This PR aims at a LocalStack-less, AWS-deployable filemanager meso-service (since it's stateful w/ database et al).

Goal for this getting merged is Victor being able to deploy/operate/develop on it from the get-go, without local errors.

Will fix issues: #70 #71  ... at least

### DIFF
--- a/lib/workload/stateful/filemanager/deploy/stack/filemanager_stack.ts
+++ b/lib/workload/stateful/filemanager/deploy/stack/filemanager_stack.ts
@@ -18,7 +18,6 @@ import * as secretsmanager from 'aws-cdk-lib/aws-secretsmanager';
 interface Settings {
   database_url: string;
   endpoint_url?: string;
-  force_path_style: boolean; //FIXME: Does this refer to buckets? We cannot do this forever as AWS will move to domain style?
   stack_name: string;
   buildEnvironment?: NodeJS.ProcessEnv;
 }
@@ -89,7 +88,6 @@ export class FilemanagerStack extends Stack {
       environment: {
         DATABASE_URL: settings.database_url,
         ...(settings.endpoint_url && { ENDPOINT_URL: settings.endpoint_url }),
-        FORCE_PATH_STYLE: settings.force_path_style.toString(),
         SQS_QUEUE_URL: queue.queueUrl,
         RUST_LOG: 'info,filemanager_ingest_lambda=trace,filemanager=trace',
       },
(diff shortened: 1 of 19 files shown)


brainstorm:
Since we removed LocalStack from the mix, the current CI will most likely fail, btw.

lib/workload/stateful/filemanager/deploy/stack/filemanager_stack.ts
```diff
@@ -4,18 +4,21 @@ import * as iam from 'aws-cdk-lib/aws-iam';
...
 
 /**
  * Common settings for the filemanager stack.
  */
 interface Settings {
   database_url: string;
-  endpoint_url: string;
-  force_path_style: boolean;
+  endpoint_url?: string;
+  force_path_style: boolean; //FIXME: Does this refer to buckets? We cannot do this forever as AWS will move to domain style?
```
mmalenic, line 21: I think you can remove any `path_style` settings because it was only was only used for localstack.

lib/workload/stateful/filemanager/deploy/stack/filemanager_stack.ts
```diff
@@ -100,6 +102,52 @@ export class FilemanagerStack extends Stack {
...
+        secretStringTemplate: JSON.stringify({ username: 'filemanager' }),
+        excludePunctuation: true,
+        generateStringKey: 'password',
+      },
+    });
+
+    // RDS
+    new rds.DatabaseCluster(this, 'Database', {
+      engine: rds.DatabaseClusterEngine.auroraPostgres({
+        version: rds.AuroraPostgresEngineVersion.VER_15_3,
```
mmalenic, line 140: Do we want to use postgres 16? Or does Aurora not support it yet?

mmalenic (approved):

lib/workload/stateful/filemanager/deploy/stack/filemanager_stack.ts
brainstorm: It was not on the enum/autocomplete 🤷🏻

brainstorm:
@mmalenic I think that you approved this PR too fast :P

Can you try to deploy yourself (against dev account) and let me know why the RDS construct is not showing up in neither cdk synth/deploy?

lib/workload/stateful/filemanager/deploy/stack/filemanager_stack.ts
mmalenic: It looks like it is supported, although only very recently (from November 15), so it's probably not in CDK yet. Should be able to set it manually though: https://github.com/aws/aws-cdk/issues/23253#issuecomment-1340935712.

mmalenic:
Sure, I'll have a look. I guess we're using Aurora over plain RDS because it's more managed/simpler?

PR #75, opened 2023-12-13
filemanager: setup testing

### LINKED ISSUE #71: filemanager: setup testing mocks

mmalenic:
Closes #71

### Changes
* Adds database and unit tests.
* Adds AWS test mocks using [mockall](https://github.com/asomers/mockall).
* General refactor to increase clarify and update docs.
* Fix ordering and equality bug when de-duplicating and sorting events within the same Lambda function.

It should be a lot easier to test locally now just by running `cargo test`. This will mean that dev deployments are more likely to succeed if the tests pass locally. Now, if only there is way to remove `docker compose up` as well. Just being able to run `cargo test` directly would be the most convenient.


lib/workload/stateful/filemanager/filemanager/src/clients/aws/config.rs
```diff
@@ -0,0 +1,26 @@
+use aws_config::{defaults, BehaviorVersion, ConfigLoader, SdkConfig};
+use mockall::automock;
+
+/// A wrapper around a config loader.
+#[derive(Debug)]
+pub struct Config {
+    inner: ConfigLoader,
+}
+
+#[automock]
```
mmalenic, line 10: This creates an entirely new `MockConfig` struct which has methods that allow setting expectations.

lib/workload/stateful/filemanager/filemanager/src/events/aws/collecter.rs
```diff
@@ -0,0 +1,246 @@
+#[double]
```
mmalenic, line 1: This expands to:
```rust
#[cfg(not(test))]
use crate::clients::aws::s3::Client;
#[cfg(test)]
use crate::clients::aws::s3::MockClient as Foo;
```
Which allows using different versions of the client in testing code compared to the main code.

brainstorm (approved):
Very nice use of mockall and mock_double, I didn't know those crates, kudos Marko!

mmalenic:
Do you want to go for #74 first? There might be some conflicts.

brainstorm:
> Do you want to go for #74 first? There might be some conflicts.

Ok, I'll put it up for review (now it's in draft), but there's still things left to fix :_S

PR #79, opened 2023-12-20
filemanager: deploy changes and fixes

mmalenic:
### Changes

* Refactored deployment into constructs:
    * Add RDS serverless v2 (you were right in using `DatabaseCluster` for this @brainstorm. https://github.com/aws/aws-cdk/issues/20197 is now fixed and cluster instances can be created with `ClusterInstance.serverlessV2`. @andrewpatto this is related to https://github.com/elsa-data/aws-infrastructure/blob/be2f0693df6bec1d90168d5441f910ce44094262/packages/stack/rds/serverless-base-database.ts#L58-L69
    * Added database migration within CDK similar to https://github.com/aws-samples/amazon-rds-init-cdk. This has the advantage of applying migrations within CDK and rejecting the deployment if it fails.
    * Make sccache optional and use env variables for fetching the binary to make it more cross platform.
* Implement `filemanager-migrate-lambda` crate for database migrations using sqlx.
* Make S3 event parsing less strict:
    * Size and e_tag should be optional. Size is not present on `s3:ObjectRemoved:*` events.
    * Avoid errors for unknown event types when deserializing.
* The default storage class when calling `HeadObject` is `Standard` if it is not specified in the output.


andrewpatto:
@mmalenic can you PR the DatabaseCluster fix onto the elsa infrastructure (if easy to do).. no rush

mmalenic:
> @mmalenic can you PR the DatabaseCluster fix onto the elsa infrastructure (if easy to do).. no rush

Yep, can do.

lib/workload/stateful/filemanager/deploy/lib/filemanager_stack.ts
```diff
@@ -0,0 +1,126 @@
...
+      enforceSSL: true,
+      removalPolicy: RemovalPolicy.DESTROY,
+    });
+    const testBucketPolicy = new PolicyStatement({
+      actions: ['s3:List*', 's3:Get*'],
+      resources: ['arn:aws:s3:::*'],
+    });
+    testBucket.addEventNotification(EventType.OBJECT_CREATED, new SqsDestination(queue));
+    testBucket.addEventNotification(EventType.OBJECT_REMOVED, new SqsDestination(queue));
+
```
andrewpatto, line 55: Not for this PR - but what will be the general story about subscribing to buckets.. presumably the file manger is not going to be responsible for creating all the BYOB buckets (I mean they probably exist completely separate to orcabus).

andrewpatto (approved):

lib/workload/stateful/filemanager/database/migrations/0001_add_object_table.sql
```diff
@@ -7,13 +7,13 @@ create table object (
     -- The name of the object.
     key varchar(1024) not null,
     -- The size of the object.
-    size int not null,
+    size int default null,
     -- A unique identifier for the object, if it is present.
     hash varchar(255) default null,
```
andrewpatto, line 12: I know it isn't in this PR - but I am a bit suspicious about this modelling.
I think you'll end up in scenarios with multiple hashes (possibly needing to know this difference between them - so you can pass on a SHA-256 hash downstream say)

andrewpatto: Also, this is a unique identifier for the "content" of the object..

lib/workload/stateful/filemanager/deploy/lib/filemanager_stack.ts
mmalenic: Yes this will need to be different in the future. I'm not sure where the best spot to put this is though. I guess from the filemanager's perspective it just needs access to the SQS queue.

lib/workload/stateful/filemanager/database/migrations/0001_add_object_table.sql
mmalenic: Yes... I'm imagining some scenario where someone wants a SHA-256 rather than just a check to see if this object is the same? I think a simple solution is to just have multiple fields for some common checksum types, e.g. SHA-256, MD5, etc (or have a type field which indicates the checksum algorithm). This also maps nicely to the [`HeadObject`](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html) outputs. Although I don't know if it's promised what kind of checksum the AWS etag is, I think it's MD5?

andrewpatto: etag is MD5 *until* the object gets above a certain size (which is definite for bioinformatics) - at which point it is not an MD5. Best to define it as its own special type "AWS-ETAG"

lib/workload/stateful/filemanager/deploy/lib/filemanager_stack.ts
andrewpatto: https://github.com/aws/aws-cdk/issues/2004

brainstorm: So @andrewpatto, if I understand that issue correctly, you'd re-define the CDK stack manually each time a new bucket is introduced?

I was thinking that perhaps getting rid altogether of S3 bucket notification definitions in CDK and do it more dynamically after the stack is deployed? Otherwise we'll have to re-deploy at each new bucket that filemanager needs to watch, not sure if that'll be practical 🤔

lib/workload/stateful/filemanager/database/migrations/0001_add_object_table.sql
brainstorm: Hm, sounds like we'll have to tackle some sort of computationally lightweight hashing scheme (via some reliably unique subsampling perhaps?)... totally part of another PR, agree.

brainstorm (approved):
Thanks for introducing the migrations as a lambda, didn't think about that at all.

Let's discuss the other couple of issues/comments from Andrew, I think they should be tackled separate and are important to get right.

PR #80, opened 2023-12-20
Gitpod

andrewpatto:
Added the files (WIP) for gitpod
Experiement starting to split the docker compose into separate units


docker-compose.yml
```diff
@@ -4,6 +4,9 @@
 
 version: '3.1'
 
+include:
+  - "mock-ica/docker-compose-mock-ica.yml"
```
andrewpatto, line 8: If we split the mocks into various categories - we can include them from the master docker compose - but also spin them indvidually as is needed

andrewpatto:
I don't know why the yarn lock file changed - I just did a yarn install.. but anyhow

victorskl:
Having a look now

docker-compose.yml
```diff
@@ -1,8 +1,10 @@
 # NOTE:
 # If you update image version, make sure to update in docker-compose.ci.yml as well as
-# also do update GHA caches step in prbuild.yml
+# also do update GHA caches step in prebuild.yml
```
victorskl, line 3: Looks like spelling correction. Hehe. While at it, could you also help update better point to the full path i.e. `.github/workflows/prbuild.yml` i.e. PullRequest Build GH workflow

docker-compose.yml
```diff
@@ -1,8 +1,10 @@
...
 
-version: '3.1'
+include:
+  - path:
+      - mock-ica/docker-compose-mock-ica.yml
```
victorskl, line 7: Oh, nice. I didn't aware of this include yet..!

.github/workflows/prbuild.yml
```diff
@@ -97,4 +97,4 @@ jobs:
 
       - name: Tear down compose stack
         run: |
-          docker-compose -f docker-compose.yml -f docker-compose.ci.yml down
+          docker compose -f docker-compose.yml -f docker-compose.ci.yml down
```
victorskl, line 100: It could be that GH runner might able to recognise subcommand `compose` (as docker cli plugin) now. Last time I tried, it failed on me. But, worth give it a shot again. So, all good and try again now.

victorskl (approved):
Look good, Patto. I will try myself GitPod later. Just one minor comment to attend. Otherwise good to go.!

docker-compose.yml
andrewpatto: Oh sorry.. my mistake

.github/workflows/prbuild.yml
andrewpatto: Yes they are deprecating the "docker-compose" - and the GH runner now understands the newer compose. Is needed to get some of the newer compose functionality ("include")

victorskl: Right! Thanks for pointer.

victorskl:
@andrewpatto FYI, I am going to merge this for now. About to rearrange setup a bit. So, will include considering this PR changes too.

andrewpatto:
Happy for this merge - actually now thinking if we do this kind of thing we should do it with devcontainers (same as gitpod - but more open standard)

Also that lets us use Github native codespaces

Issue #73, opened 2023-12-11
filemanager: ensure S3 events are in order


mmalenic:
Similar to #72, events received from AWS S3 can be out of order. This is important if receiving an `s3:ObjectRemoved:*` event before an `s3:ObjectCreated:*` when the created event was meant to come first. In order to ensure a consistent state, this scenario should be accounted for. One method to do this is to track the `sequencer` [field](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html) which indicates the correct order of events for the same key and bucket. On inserting events into the database, this value can be queryed to find if a `s3:ObjectRemoved:*` event has a corresponding  `s3:ObjectCreated:*`.

PR #88, opened 2024-01-23
feat: de-duplicate events at the database level

### LINKED ISSUE #72: filemanager: de-duplicate S3 event messages
Similar to #73, it is possible to receive duplicate events from AWS S3 notifications. At the database level, or via some other mechanism, events should be de-duplicated to ensure that a consistent state is maintained. One method to do this is to track incoming events and ensure that no duplicate events get ingested to the database using `on conflict do nothing` when [inserting](https://www.postgresql.org/docs/current/sql-insert.html) events.

mmalenic:
Closes #72

This was more complex than I expected, especially because there is a lot of caveats in postgres related to upserts and concurrency errors. I think the `on conflict` solution works well to avoid these.

Ignore the commits related to macros, I was experimenting with writing postgres functions in Rust using [plrust](https://github.com/tcdi/plrust) and converting regular functions in code into plrust functions using attribute macros. However, I don't think this is useful for this PR.

### Changes
* Introduce sequencer values and version ids into the s3 object table so that duplicate events can be ignored using `insert ... on conflict`.
    * `on conflict` avoids any concurrency issues related to manually upserting objects.
    * `bucket` and `key` are pushed from `object` to `s3_object`. This helps with de-duplication, and also makes more sense if considering an object which is the same but living in two different places.
    * For now, `checksum` and `size` are in `object`, but this could be changed if two objects that are the same have different checksums, for example, different kinds of compression. If this this case, I think it would make sense to have another table one level up which links physical objects with the same checksums/sizes to `logical` or `semantic` objects.
* This de-duplication logic is mirrored in the code within one Lambda call. This logic is the same as described [here](https://github.com/aws-samples/amazon-s3-endedupe/tree/bd906412c2b4ca26eee6312e3ac99120790b9de9), where events with identical `key`, `bucket`, `version_id` and `sequencer` values are considered duplicates.
* Add various tests to verify de-duplication.


lib/workload/stateful/filemanager/database/queries/ingester/aws/insert_s3_deleted_objects.sql
```diff
@@ -0,0 +1,31 @@
...
+   unnest($5::timestamptz[]),
+   unnest($6::timestamptz[]),
+   unnest($7::timestamptz[]),
+   unnest($8::text[]),
+   unnest($9::storage_class[]),
+   unnest($10::text[]),
+   unnest($11::text[])
+) on conflict on constraint deleted_sequencer_unique do update
+    set number_duplicate_events = s3_object.number_duplicate_events + 1
+    returning object_id, number_duplicate_events;
```
mmalenic, line 31: This isn't being used yet, however I think it will be for #73.

brainstorm (approved):

PR #91, opened 2024-01-29
Update RDS Construct

williamputraintan:
Mainly the switch from `mysql` to `postgres` RDS

### DIFF
--- a/lib/workload/stateful/database/component.ts
+++ b/lib/workload/stateful/database/component.ts
@@ -1,9 +1,31 @@
 import { Construct } from 'constructs';
-import { RemovalPolicy } from 'aws-cdk-lib';
+import { RemovalPolicy, Duration } from 'aws-cdk-lib';
 import * as rds from 'aws-cdk-lib/aws-rds';
 import * as ec2 from 'aws-cdk-lib/aws-ec2';
 
-export interface DatabaseProps {
+/**
+ * Props for enabling enhanced monitoring.
+ */
+type MonitoringProps = {
+  /**
+   * Add cloud watch exports.
+   */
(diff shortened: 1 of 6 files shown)


lib/workload/stateful/database/component.ts
```diff
@@ -1,86 +1,96 @@
...
 export class DatabaseConstruct extends Construct {
   readonly dbSecurityGroup: ec2.SecurityGroup;
   readonly dbCluster: rds.DatabaseCluster;
 
   constructor(scope: Construct, id: string, vpc: ec2.IVpc, props: DatabaseProps) {
     super(scope, id);
-    const dbPort = 3306;
 
-    const secret = new rds.DatabaseSecret(this, id + 'Secret', {
+    const dbSecret = new rds.DatabaseSecret(this, id + 'DbSecret', {
```
mmalenic, line 48: Is the idea that every microservice that needs the database will fetch this secret to connect?

mmalenic (approved):

lib/workload/stateful/database/component.ts
williamputraintan: I am thinking this way but do you have any suggestion?

Realising this, I should have pass in the secret name so that it could be discovered by the microservice.

mmalenic: I think this is okay. Probably simplest to just have one secret for the whole database cluster.

reisingerf: Ha, I was thinking similarly for the DB stack to set up databases / users for the different services and isolate them against each other... but let's start simple ;-)

PR #92, opened 2024-02-02
Initiate CDK Pipeline

williamputraintan:
- Deployed 2 CDK pipelines (stateful and stateless) at the bastion account
- Added cdk test with [cdk-nag](https://github.com/cdklabs/cdk-nag)


bin/stateless-pipeline.ts
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+import 'source-map-support/register';
+
+import * as cdk from 'aws-cdk-lib';
+import { StatelessPipelineStack } from '../lib/pipeline/orcabus-stateless-pipeline-stack';
+
+const AWS_TOOLCHAIN_ACCOUNT = '383856791668'; // Bastion
```
brainstorm, line 7: Why the Bastion account?

brainstorm (approved):
Nice, first time I see `cdk-nag` being used!

Where can I see a successful run of this? Bastion (why Bastion, btw)? I'm interested in runtime and whether the outputs are easily visible without auth (publicly and safely, like GHA).

victorskl:
I will review in tick, soon

bin/stateless-pipeline.ts
victorskl: We opt to CICD within the CodePipeline. So, the GHA PR build will only do pre-flight checks - lint, format, secret-scan, dependencies audits.

Choosing the Bastion account because we still opt with our AWS accounts. There, Bastion account has been repurposed as build/toolchain account.

victorskl:
[OrcaBus V1: Setup test framework for runtime and deployment CDK Constructs](https://trello.com/c/4b8Nmgrb/1444-orcabus-v1-setup-test-framework-for-runtime-and-deployment-cdk-constructs)

victorskl:
[OrcaBus V1: Implement baseline CDK Pipeline](https://trello.com/c/7ZOhDVFe/1438-orcabus-v1-implement-baseline-cdk-pipeline)

lib/workload/stateless/sequence_run_manager/Makefile
```diff
@@ -11,19 +11,14 @@ lint:
...
-	@sleep 8
+# full mock suite test pipeline - install deps, bring up compose stack, run suite, bring down compose stack
+test: install up suite down
 
 suite:
 	@python manage.py test
 
 up:
-	@docker compose up -d
+	@docker compose up --wait -d
```
victorskl, line 21: Oh, nice! Thank you.

lib/workload/stateless/metadata_manager/package.json
```diff
@@ -36,6 +36,7 @@
     "@types/lodash": "^4.14.202",
     "@types/node": "^20.11.5",
     "jest": "^29.7.0",
+    "jest-junit": "^16.0.0",
```
victorskl, line 39: 😄 Ah, so jest uses junit format for reporting. Good to know.!

package.json
```diff
@@ -12,23 +12,25 @@
...
     "clean": "tsc --build --clean",
     "audit": "yarn npm audit",
     "lint": "eslint .",
     "lint-fix": "eslint --fix .",
     "prettier": "prettier --check .",
-    "prettier-fix": "prettier --write ."
+    "prettier-fix": "prettier --write .",
+    "cdk-stateless-pipeline": "cdk --app 'yarn run -B ts-node --prefer-ts-exts bin/stateless-pipeline.ts'",
+    "cdk-stateful-pipeline": "cdk --app 'yarn run -B ts-node --prefer-ts-exts bin/stateful-pipeline.ts'",
+    "cdk-orcabus": "cdk --app 'yarn run -B ts-node --prefer-ts-exts bin/orcabus.ts'"
```
victorskl, line 25: Could you pls reflect these changes to top level README, too?

package.json
```diff
@@ -12,23 +12,25 @@
   "bin": "bin/pipeline.js",
```
victorskl, line 12: Perhaps, also update this bin pointer to `bin/orcabus.ts` as default.

config/constants.ts
```diff
@@ -94,6 +94,7 @@ export const getEnvironmentConfig = (
               maxACU: 1,
```
victorskl, line 94: Wouldn't mind set this up to 16 ACU as it is serverless ops. It will scale automatically down 0.5 when no load. FM could easily hit 1 ACU limit.

williamputraintan:
> Where can I see a successful run of this?
Just updated and could access it in the codebuild report tab. Or could navigate from the pipeline: `(UnitTest) Aws CodeBuild` -> `Report`.

Quick link:
- [stateless-pipeline](https://ap-southeast-2.console.aws.amazon.com/codesuite/codebuild/383856791668/projects/PipelineBuildUnitTestF33196-g8tv44iMvGxB/build/PipelineBuildUnitTestF33196-g8tv44iMvGxB%3Aa74c96fa-d56f-46df-bce3-2ba70355ae8b/reports?region=ap-southeast-2&reports-meta=eyJmIjp7InRleHQiOiIifSwicyI6eyJwcm9wZXJ0eSI6InN0YXR1cyJ9LCJuIjoyMCwiaSI6MH0)
- [stateful-pipeline](https://ap-southeast-2.console.aws.amazon.com/codesuite/codebuild/383856791668/projects/PipelineBuildUnitTestF33196-p7ubX5tBDbyQ/build/PipelineBuildUnitTestF33196-p7ubX5tBDbyQ%3A80df2ef9-c04d-499a-a2c5-43cd508aadf2/reports?region=ap-southeast-2&reports-meta=eyJmIjp7InRleHQiOiIifSwicyI6eyJwcm9wZXJ0eSI6InN0YXR1cyJ9LCJuIjoyMCwiaSI6MH0)

The last pipeline run is failing because it listen to `main` branch and this has not been merged yet.

lib/workload/stateless/sequence_run_manager/Makefile
williamputraintan, line 21: @victorskl This `--wait` flag should do the wait for the database up. Added a startup period on the docker-compose for the healthcheck

config/constants.ts
```diff
@@ -24,7 +24,7 @@ const orcaBusStatefulConfig = {
     defaultDatabaseName: 'orcabus',
     version: AuroraPostgresEngineVersion.VER_15_4,
     parameterGroupName: 'default.aurora-postgresql15',
-    username: 'admin',
+    username: 'postgres',
     dbPort: 5432,
     masterSecretName: rdsMasterSecretName,
```
victorskl, line 29: We have to create per app/service db secret manager record per database to isolate them better. Instead of reusing this master (SA account). Wonder, we can do this arrangement straight up now at this point... By chance, do you happen to figure any solution to it?

victorskl (approved):
Looking good, Will. Great job setting this up! Few minor comments to attend and/or to discuss further. Approve otherwise.

config/constants.ts
reisingerf: I agree that we should isolate on a per service level, but that means the stateless stack needs to be aware of the consuming services, right? 
And any database required by any service would have to be registered to and created by the stateless stack IIRC. The stateless stack can create "user" credentials for each service, but I think they'd need to be linked to existing DBs? I don't have the understanding yet of what's possible...

I think the initial thinking here was to allow access on the cluster level to start with in order to not block any service development. Once we've figured out a better way, we could refactor. 

Or do you think it's worth working that out now and safe on the refactor ?

victorskl: I'd recommend we should work this out now.

mmalenic: I agree in working this out. Could be wrong, but it doesn't look like CDK can create more than one database/user per cluster using the regular higher level constructs (see issue https://github.com/aws/aws-cdk/issues/13588)? Might have to create a migration-style Lambda function which executes `create database ...` after the RDS cluster is up and running, e.g. using something like [this](https://github.com/umccr/orcabus/blob/50a6f662ab8723ad3a71595a6ef3f84a1a70f192/lib/workload/stateful/filemanager/deploy/constructs/cdk_resource_invoke.ts#L44). Then, I think each service would need to register their preferred database name, user and credentials somewhere (in constants.ts?).

mmalenic: Or, each service could register their own migration function to create the databases/users? That way the databases wouldn't have to be registered somewhere higher up like in `constants.ts`. Not sure if this makes more sense?

reisingerf: Hm, I wonder if the stateless stack could create IAM roles/policies for each service that would allow these services access to only their specific DB(s) on the instance. E.g. passing on the permissions to do what the service needs to do (without control/knowledge of what that may be), as long as it happens in the designated DB. 

I haven't found any examples, but it looks like you can create IAM policies with conditions to restrict access by [rds:DatabaseName](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/security_iam_service-with-iam.html#UsingWithRDS.IAM.Conditions).

package.json
williamputraintan: I think we could remove this bin

config/constants.ts
williamputraintan: Will track this at issue #99

bin/stateless-pipeline.ts
victorskl: As discussed; will reenable GHA PR Build with #100

victorskl:
Folks, would to merge this sometime today pls. Hope we are all good, if no more comments, hereafter.

reisingerf (approved):

victorskl:
@mmalenic Marrrko; is this ok on your side? Planning to merge this in tick, next.  Hopefully, you can rebase onto it once merge to main.

mmalenic:
Yes, go for it.

mmalenic (approved):

bin/stateless-pipeline.ts
victorskl: Also will tear down and re-deploy into UoM AWS accounts with #102

victorskl:
Merging...

PR #93, opened 2024-02-05
feat(filemanager): ensure events are ingested in the correct order

### LINKED ISSUE #73: filemanager: ensure S3 events are in order

mmalenic:
Closes #73 

There's still a bit of repetitive code in the application logic and queries, but I think for now it's okay. It will be interesting to benchmark these queries eventually, especially within the `update_reordered_for_*.sql` queries because the select statement locks all rows with the same bucket, key and version_id. The `update_reordered_for_*.sql` queries are a bit long, but I think it's okay because it reduces the number of requests that the Lambda function needs to make to the database. Potentially the update and insert queries could be merged to reduce code repetition, with a flag variable indicating if it is a created or deleted event.

### Changes
* Make sure that events are ingested in the correct order by using the `created_sequencer` and `deleted_sequencer` values in the database table. The process is:
   * For each event, check if the sequencer condition is met, where there is a sequencer value that better matches any already ingested event with the same bucket, key and version id.
   * Then, replace the already ingested event with the new one, and return the old event to be re-ingested, as it could belong to another object.
   * If there is no out of order event, then just insert the event normally.
   * This process allows the database to correct any ordering issues as it receives events, where the current state of the database represents the best known order of all events received.
 * Move the size and checksum from `object` to `s3_object` because this simplifies the reordering logic.
 * Add various tests:
    * More ordering and duplication tests.
    * Tests for the queries themselves (rather than inside the ingest function).
    * Long permutation test to make sure event reordering works correctly. This could also be rewritten as a benchmark or performance test.


victorskl:
I will review in tick, soon

lib/workload/stateful/filemanager/database/queries/ingester/aws/update_reordered_for_created.sql
```diff
@@ -0,0 +1,118 @@
...
+            -- If a sequencer already exists this event should be reprocessed because this
+            -- sequencer could belong to another object.
+            current_objects.created_sequencer < current_objects.input_created_sequencer
+        )
+        -- And there should not be any objects with a created sequencer that is the same as the input created
+        -- sequencer because this is a duplicate event that would cause a constraint error in the update.
+        and current_objects.input_created_sequencer not in (
+            select created_sequencer from current_objects where created_sequencer is not null
+        )
+),
```
mmalenic, line 79: Given that this is a bit of a monster statement. I think the understandability would be greatly improved if this was split across multiple SQL queries, where the application code did the majority of this sequencer logic/comparison. However, this would be at the cost of more application <-> database calls, although I'm not sure if this matters that much for now.

lib/workload/stateful/filemanager/filemanager/README.md
```diff
@@ -20,12 +20,45 @@ contain thin wrappers around clients, and any logic should be elsewhere to ensur
...
 
+Generally code that belongs to a particular cloud service should be put in its own module (e.g. AWS code goes in an `aws` module).
+
+## Event ingestion
+
+The filemanager determines the location of objects in cloud storage by ingesting events like the [AWS S3 events][s3-events].
+However, events for most cloud services only guarantee that events are received at least once. Duplicate events may be received
+or events could be received out of order, and the filemanager should handle this. This needs to occur both in application
+code and at the database level because events can be out of order across Lambda function calls. The aim of the filemanager
+is to maintain a database state that is as correct as possible at the time an event is received.
```
mmalenic, line 31: Currently the aim is to keep the database as up to date as possible, as soon as an event is received. However, as @brainstorm suggested, it might be better to keep events ordered within a sliding window of time, and wait a certain period (i.e. 5 mins, 1hr, etc) before inserting.

I'm not sure if AWS guarantees anything in terms of events arriving out of order within a certain time period, although in practice I don't think this window would be huge?

lib/workload/stateful/filemanager/database/queries/ingester/aws/update_reordered_for_created.sql
brainstorm: Doesn't look that bad? The comments make it more understandable... how large would the perf impact be if you split?

mmalenic: I can't image this cost would be that great. It's not like it's an order of magnitude thing, I think it would be down to how well the database could optimize the bigger query.

brainstorm: Up to you if you want to refactor, I'd leave it as-is... it's more readable today than it was yesterday (to me) so this piece of code complexity is a bit subjective in this case, I just needed to be in the right frame of mind to understand it and its context :-S

That being said, I think that an ARCHITECTURE.md explanation would be helpful regardless: https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html

lib/workload/stateful/filemanager/filemanager/README.md
brainstorm: Worth checking the SQS docs, defaults and guarantees on that.

lib/workload/stateful/filemanager/filemanager/src/events/aws/collecter.rs
brainstorm, line 1: collecter.rs->collector.rs?

lib/workload/stateful/filemanager/filemanager/src/events/aws/collector_builder.rs
```diff
@@ -1,3 +1,6 @@
+use mockall_double::double;
```
brainstorm, line 1: Why is this not part of collector.rs instead of a separate file?

lib/workload/stateful/filemanager/filemanager/src/events/aws/mod.rs
```diff
@@ -144,11 +140,62 @@ impl From<FlatS3EventMessages> for TransposedS3EventMessages {
     }
 }
 
+impl From<TransposedS3EventMessages> for FlatS3EventMessages {
```
brainstorm, line 143: This definitely needs a comment, I always forget why the flattening and transposing is needed :-S

brainstorm (changes requested):

lib/workload/stateful/filemanager/filemanager/src/events/aws/mod.rs
mmalenic: Yes, fair. The transpose is just to help the shape of the query entering the database. I.e. using `unnest` on the input parameters.

lib/workload/stateful/filemanager/filemanager/src/events/aws/collector_builder.rs
mmalenic: Good point, could be the same file.

lib/workload/stateful/filemanager/filemanager/src/events/aws/collecter.rs
mmalenic: Hmm.. I went with `ingester` instead of `ingestor`, because it.. sounded better?

victorskl (approved):
Very detailed impl, Marko. Kudos.! Can't wait to see this in action.

brainstorm (approved):

PR #98, opened 2024-02-08
refactor: filemanager deployment code and Makefile

### LINKED ISSUE #85: filemanager: refactor code to match project structure
The filemanager code should be refactored to match the project structure. This includes:
* Removing stateful CDK code from the filemanager and using the shared VPC, database, etc.
* Removing bucket creation in the filemanager stack.
* Moving filemanager from the stateful directory to the stateless directory.
* Adding a `Makefile` that can be used with the outer `Makefile` including install, clean, test, etc. This is related to https://github.com/umccr/orcabus/issues/76.

mmalenic:
Closes #85.

No rush on this, I can merge after https://github.com/umccr/orcabus/pull/92.

### Changes
* Add Makefile consistent with other services that defines build, test, lint, etc.
* Remove stateful constructs from the filemanager and use the shared database cluster.
* Move filemanager from the stateful to stateless directory.
* Make `cdk_resource_invoke.ts` generic over the type of Lambda function so that it can be used in other contexts and not just Rust functions.
    * I feel like this could be useful for other services, e.g. creating multiple databases within one database cluster (related to https://github.com/umccr/orcabus/pull/92).


lib/workload/orcabus-stateless-stack.ts
```diff
@@ -41,10 +63,50 @@ export class OrcaBusStatelessStack extends cdk.Stack {
...
+    }
   }
 
   private createSequenceRunManager() {
     // TODO new SequenceRunManagerConstruct() from lib/workload/stateless/sequence_run_manager/deploy/component.ts
     //   However, the implementation is still incomplete...
   }
+
+  private createFilemanager(dependencies: FilemanagerDependencies, vpc: IVpc) {
+    // Opting to reconstruct the dependencies here, and pass them into the service as constructs.
```
mmalenic, line 78: I have a feeling this pattern will be very common, so it might be good to have another class somewhere that does this exclusively, e.g. `DependencyReconstructor.ts`.

lib/workload/stateful/database/component.ts
```diff
@@ -16,59 +18,120 @@ type MonitoringProps = {
...
+      allowAllIpv6Outbound: false,
+      securityGroupName: props.inboundSecurityGroupName,
+      description: 'an inbound security group to connect to the OrcaBus RDS',
+    });
+
+    this.securityGroup.addIngressRule(
+      this.inboundSecurityGroup,
+      ec2.Port.tcp(props.dbPort),
+      'allow the OrcaBus security group to access db'
+    );
```
mmalenic, line 132: I'm not sure exactly what the best approach is with these security groups, but I figure there should be an "inbound" security group which the stateless can lookup.

lib/workload/stateless/filemanager/.sqlx/query-6515aa29536fb79eb777151f8650c8eec41a9bd4bcc72a06d36a5d0eed482115.json
```diff
@@ -0,0 +1,117 @@
+{
```
mmalenic, line 1: FYI, these `query-*.json` files are required for filemanager to compile without an active postgres database running, which means that any `cdk synth` doesn't need to know about the filemanager's docker compose configuration.

mmalenic (commented):
Should be ready for a review now, /cc @victorskl.

lib/workload/orcabus-stateless-stack.ts
```diff
@@ -12,7 +15,26 @@ export interface OrcaBusStatelessConfig {
...
+   */
+  eventSourceBuckets: string[];
+  /**
+   * Database secret name for the filemanager.
+   */
+  databaseSecretName: string;
+  /**
+   * Database security group name to allow the filemanager Lambda to connect.
+   */
+  databaseSecurityGroupName: string;
```
victorskl, line 37: Could you reuse `lambdaSecurityGroupName` from base stateless config? That was the intent as shared SG.

lib/workload/orcabus-stateless-stack.ts
```diff
@@ -41,10 +63,50 @@ export class OrcaBusStatelessStack extends cdk.Stack {
...
+        },
+        this
+      )
+    );
+    const databaseSecurityGroup = SecurityGroup.fromLookupByName(
+      this,
+      'FilemanagerDatabaseSecurityGroup',
+      dependencies.databaseSecurityGroupName,
+      vpc
+    );
```
victorskl, line 95: Similarly, you could reuse (which is commented around line 49-56) instead. You may then reuse along like `this.vpc` as class property.

lib/workload/orcabus-stateless-stack.ts
```diff
@@ -41,10 +63,50 @@ export class OrcaBusStatelessStack extends cdk.Stack {
...
+
+    new Filemanager(this, 'Filemanager', {
+      buckets: dependencies.eventSourceBuckets,
+      buildEnvironment: {},
+      databaseSecret,
+      databaseSecurityGroup,
+      eventSources: [queue],
+      migrateDatabase: true,
+      vpc: vpc,
```
victorskl, line 109: Simply `vpc: this.vpc` perhaps? Instead of passing in...

lib/workload/stateful/database/component.ts
```diff
@@ -16,59 +18,120 @@ type MonitoringProps = {
...
+   */
   parameterGroupName: string;
+  /**
+   * Database username.
+   */
   username: string;
+  /**
+   * Database security group name.
+   */
+  securityGroupName: string;
```
victorskl, line 53: This will be internal to database itself. We won't need/want to expose it. Or, let external construct control/manipulate it. Hence, it is not configurable or, need not to be configured.

lib/workload/stateful/database/component.ts
```diff
@@ -16,59 +18,120 @@ type MonitoringProps = {
...
+  /**
+   * The database removal policy.
+   */
   removalPolicy: RemovalPolicy;
-  allowedInboundSG?: ec2.SecurityGroup;
+  /**
+   * Create an inbound security group that can connect to the database. Stateless resources can recreate
+   * this security group to access the database.
+   */
+  inboundSecurityGroupName: string;
```
victorskl, line 86: Same concept; the inversion of control (IoC) here. 

Thinking of like; if you pass me along inbound SG (i.e. optional `allowedInboundSG?` ts trick) during db construction time then I will add this SG as source inbound allow rule to my private internal security.

config/constants.ts
```diff
@@ -28,9 +31,11 @@ const orcaBusStatefulConfig = {
     username: 'postgres',
     dbPort: 5432,
     masterSecretName: rdsMasterSecretName,
+    securityGroupName: 'orcabus-database-security-group',
```
victorskl, line 34: As explained, we'd not need this expose.

config/constants.ts
```diff
@@ -28,9 +31,11 @@ const orcaBusStatefulConfig = {
...
     monitoring: {
       cloudwatchLogsExports: ['orcabus-postgresql'],
     },
+    inboundSecurityGroupName: 'inbound-database-security-group',
```
victorskl, line 38: ditto

config/constants.ts
```diff
@@ -28,9 +31,11 @@ const orcaBusStatefulConfig = {
...
   },
   securityGroupProps: {
```
victorskl, line 40: Here is shared Lambda/Compute SG props contant

lib/workload/stateless/filemanager/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    build: database
+    container_name: filemanager_db
+    restart: always
+    environment:
+      - POSTGRES_DATABASE=filemanager
+      - POSTGRES_USER=filemanager
+      - POSTGRES_PASSWORD=filemanager
+      - PGPORT=4321
+    ports:
+      - "4321:4321"
```
victorskl, line 14: We tend to use just `compose.yml` these days. But, that's all good.! Just saying..

test/stateless/cdkResourceInvoke.test.ts
```diff
@@ -0,0 +1,72 @@
...
+            'Fn::Join': [
+              '',
+              Match.arrayWith([`:function:${expectedHash}-ResourceInvokeFunction-*`]),
+            ],
+          },
+        },
+      ],
+    },
+  });
+});
```
victorskl, line 72: Nice! thanks for testing. :)

test/stateful/databaseConstruct.test.ts
```diff
@@ -38,17 +39,19 @@ test('Test DBCluster created props', () => {
...
-  const sgLogicalId = stack.getLogicalId(allowedSG.node.defaultChild as ec2.CfnSecurityGroup);
 
-  new DatabaseConstruct(stack, 'TestDatabaseConstruct', vpc, {
-    ...constructConfig.stackProps.orcaBusStatefulConfig.databaseProps,
-    allowedInboundSG: allowedSG,
-  });
+  const sgLogicalId = stack.getLogicalId(
+    database.inboundSecurityGroup.node.defaultChild as ec2.CfnSecurityGroup
+  );
+
```
victorskl, line 52: As explained, might want to revert/retain this.

victorskl (commented):
Looking good most of it. Except the db SG refactor may need revert as explain its original intent. Ping me pls if we'd like catchup a bit more focus/details, etc.

lib/workload/stateful/database/component.ts
mmalenic: Ahh I see, that makes a lot of sense. Thanks! So just confirming, the idea behind the `SecurityGroupConstruct` is that stateless constructs will use this to connect to the database?

E.g. for filemanager, it will lookup `SecurityGroupConstruct` using the `securityGroupName`, and then pass this into the Lambda function's security groups [here](https://github.com/umccr/orcabus/blob/360b2b3efcd347e3297a04e2381a20b8cc06deee/lib/workload/stateless/filemanager/deploy/constructs/functions/function.ts#L116-L117)?

lib/workload/stateless/filemanager/docker-compose.yml
mmalenic: Didn't know that! Seems to be change from 3 years ago https://github.com/docker/compose/releases/tag/1.28.6/.

mmalenic:
Thanks for the review @victorskl. I'll revert that SG change and also address the other recommendations.

lib/workload/stateful/database/component.ts
victorskl: Correct, yes.!

victorskl:
Sound good, no worries.

victorskl (approved):
LGTM!

PR #101, opened 2024-02-09
feat(filemanager): move SQS and DLQ to stateful stack

### LINKED ISSUE #86: filemanager: link bucket EventBridge rules to an SQS queue
An EventBridge rule should be setup on a bucket that passes events to an SQS queue that filemanager can use. This should be a separate component in the stateful directory. As part of this, a DLQ should be setup for events that cannot be processed, which triggers a CloudWatch alarm.

mmalenic:
Closes #86 

I think this one I'll merge before https://github.com/umccr/orcabus/pull/98, so that https://github.com/umccr/orcabus/pull/98 can implement the `createFilemanager()` function inside the stateless stack.

Note that the architecture is slightly different to the diagram. The Lambda function depends on an `IQueue` construct directly, rather than an SQS url, and has the option to send events to the DLQ. ~~So the final diagram could look more like this, where the DLQ receives events from both the SQS queue and the filemanager Lambda function:~~ The diagram is now like the original design.

### Changes
* Implement the EventSource construct, which contains an EventBridge rule matching S3 buckets that pushes events to an SQS queue.
    * Add a DLQ with a CloudWatch alarm that triggers when any events go to the DLQ.
* Make the stateless stack depend on the stateful stack, e.g. see https://blog.serverlessadvocate.com/serverless-aws-cdk-pipeline-best-practices-patterns-part-1-ab80962f109d#1913


config/constants.ts
```diff
@@ -98,6 +97,9 @@ export const getEnvironmentConfig = (
             securityGroupProps: {
               ...orcaBusStatefulConfig.securityGroupProps,
             },
+            eventSourceProps: {
```
brainstorm, line 100: IIRC the event sources will eventually have to be dynamic? As in not requiring a de-deploy when we want to monitor and ingest new buckets... I guess that's a future feature, but just keeping tabs open on the topic as I guess that this will eventually have to go to some Rust AWS SDK lambda instead of CDK? ... thinking out loud, no need to address this in this PR.

brainstorm (approved):

bin/orcabus.ts
```diff
@@ -27,12 +27,13 @@ const props: cdk.StackProps = {
...
 if (!config) throw new Error('No Config');
 
-new OrcaBusStatefulStack(app, 'OrcaBusStatefulStack', {
+const statefulStack = new OrcaBusStatefulStack(app, 'OrcaBusStatefulStack', {
   ...config.stackProps.orcaBusStatefulConfig,
   ...props,
 });
 
 new OrcaBusStatelessStack(app, 'OrcaBusStatelessStack', {
+  statefulStack,
```
reisingerf, line 36: The stateless stack should not depend on the stateful stack. 
Ideally there are only static parameters passed between those stacks to maintain independence. 

If I see this right, the only stateful resource that needs to be passed to the stateless stack is the SQS Queue, which should be possibly via it's ARN, no?

reisingerf (changes requested):

bin/orcabus.ts
mmalenic: Yes, could de-couple this more. An ARN should be enough to reconstruct the `IQueue` construct using [`fromQueueArn`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_sqs.Queue.html#static-fromwbrqueuewbrarnscope-id-queuearn).

Although isn't this still the stateless stack depending on the stateful stack, just more indirected? Is the advantage that it isn't the whole stateful stack, just the required resource?

reisingerf: It's less a full stack vs a single resource thing, it's more that there is now a dependency between those stacks that potentially can cause issues down the road.
(also see: https://www.endoflineblog.com/cdk-tips-03-how-to-unblock-cross-stack-references)

Also, with this setup you (in this case unnecessarily) force the stateless stack to depend on another CDK stack/construct. What if I want to create the SQS via Terraform? 

Ideally, the stateless stack should just take a list of (string) parameters and be agnostic to how those stateful resources are created/managed.

mmalenic: Yes, that's a good point. So in this case, the string parameter would be an ARN for the SQS queue, right?

config/constants.ts
mmalenic: Hmm, could potentially have env variables that contain an SQS url that the Lambda function uses. Although, wouldn't every new bucket need to somehow be registered with the SQS queue, requiring CDK? Not sure how SQS could discover these on its own, unless there's just a blanket "all buckets" EventBridge rule.

brainstorm: Exactly, @reisingerf, what were your thoughts from previous meetings on this? I'm not sure what are the requirements for this.

reisingerf: The idea is that buckets of interest are configured to send events to the EventBus, to make those discoverable. 

Marko's statefull stack will contain (and be expanded with) Event Rules that then funnel events from those buckets of interest to the SQS queue. 
The stateless FM stack does not need to change at all for this, as it simply consumes events from the SQS. 

Does that make sense?

brainstorm: Ok, so the dev story of adding a new "bucket of interest" is declaring it statically in CDK (on the EventBus construct/stack in this case) or those Event Rules will be lax enough to just process events from a new bucket? Or some other mechanism I'm not understanding? :-S

reisingerf: "adding a new bucket" is simply adding a new Rule to the stateless stack to forward the events to SQS. This should always be possible, even if there is no such bucket (as it's a simple pattern matching on the EventBus side). 

How actual events get to the EventBus is a different story and outside the scope of the stack. This would be the responsibility of whoever controls the bucket in question.

mmalenic: Could be nice to have the ability to add multiple EventBridge rules to the same SQS queue within this PR. Not sure if this is super useful though because multiple buckets can already be defined for the same rule.

reisingerf: Hm, about that... 
I need to brush up on the Rule syntax, but if I understand the code correctly then the rule filters for a known list of buckets and a given prefix, right? 

That prefix would have to be the same across all the listed buckets though? 
If yes, then I think I'd prefer a rule per bucket / concern... easier to keep track of and maintain independence.

mmalenic: Yes, it's one prefix for all the buckets. I think you're right, it might make more sense to have it defined per bucket.

lib/workload/orcabus-stateless-stack.ts
```diff
@@ -14,9 +14,37 @@ export interface OrcaBusStatelessConfig {
...
+ */
+export type EventSourceDependency = {
+  /**
+   * The SQS queue ARN.
+   */
+  queueArn: string;
+  /**
+   * The dead letter queue ARN.
+   */
+  deadLetterQueueArn: string;
```
reisingerf, line 28: Why do you want to pass on the DLQ?

mmalenic:
### Updates
* Queue name is passed in static config so that the stateless and stateful pipeline can be separate.
* EventBridges rules are configurable per bucket instead of for all buckets.
* Dead letter queue is only applied to the SQS queue within the `EventSource` construct.

test/stateful/eventSourceConstruct.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    },
+  });
+}
+
+beforeEach(() => {
+  stack = new cdk.Stack();
+});
+
+test('Test EventSource created props', () => {
+  new EventSource(stack, 'TestDatabaseConstruct', {
```
reisingerf, line 41: I guess `TestDatabaseConstruct` should be updated to `TestEventSourceConstruct`... 
(also for the following tests)

reisingerf (approved):
LGTM

PR #112, opened 2024-02-21
fix: event source name rule should be an array

### LINKED ISSUE #111: event source rules are not valid
The event source rule cannot be deployed as it is not valid. Specifically, the [`name`](https://github.com/umccr/orcabus/blob/88b5af3e5183ae5027839d7789d266631afb177a/lib/workload/stateful/event_source/component.ts#L75C17-L75C21) parameter should be an object or an array. In this case, it could just be an array with a single bucket.

mmalenic:
Closes #111 

### Changes
* Convert `EventSource` name rule to array.

### DIFF
--- a/lib/workload/stateful/event_source/component.ts
+++ b/lib/workload/stateful/event_source/component.ts
@@ -72,7 +72,7 @@ export class EventSource extends Construct {
           detail: {
             ...(prop.bucket && {
               bucket: {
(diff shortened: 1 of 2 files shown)


williamputraintan (approved):

victorskl (approved):
Yup! (Flo and me found the other day. Forget to mention...)

PR #121, opened 2024-02-26
fix(filemanager): deserialize event bridge messages

### LINKED ISSUE #119: filemanager: the structure of `S3EventMessage` is incorrect for EventBridge events
The structure of the [`S3EventMessage`](https://github.com/umccr/orcabus/blob/6fa4d2b7895310ec81353ff3db1e0bb3f5694727/lib/workload/stateless/filemanager/filemanager/src/events/aws/mod.rs#L493-L499) is only valid for messages received directly from an SQS queue, and not messages routed through EventBridge. E.g. compare the messages from SQS [directly](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html):

```json
{  
   "Records":[  
      {  
         "eventVersion":"2.2",
         "eventSource":"aws:s3",
         "awsRegion":"us-west-2",
         "eventTime":"The time, in ISO-8601 format, for example, 1970-01-01T00:00:00.000Z, when Amazon S3 finished processing the request",
         "eventName":"event-type",
         "userIdentity":{  
            "principalId":"Amazon-customer-ID-of-the-user-who-caused-the-event"
         },
         "requestParameters":{  
            "sourceIPAddress":"ip-address-where-request-came-from"
         },
         "responseElements":{  
            "x-amz-request-id":"Amazon S3 generated request ID",
            "x-amz-id-2":"Amazon S3 host that processed the request"
         },
         "s3":{  
            "s3SchemaVersion":"1.0",
            "configurationId":"ID found in the bucket notification configuration",
            "bucket":{  
               "name":"bucket-name",
               "ownerIdentity":{  
                  "principalId":"Amazon-customer-ID-of-the-bucket-owner"
               },
               "arn":"bucket-ARN"
            },
            "object":{  
               "key":"object-key",
               "size":"object-size in bytes",
               "eTag":"object eTag",
               "versionId":"object version if bucket is versioning-enabled, otherwise null",
               "sequencer": "a string representation of a hexadecimal value used to determine event sequence, only used with PUTs and DELETEs"
            }
         },
         "glacierEventData": {
            "restoreEventData": {
               "lifecycleRestorationExpiryTime": "The time, in ISO-8601 format, for example, 1970-01-01T00:00:00.000Z, of Restore Expiry",
               "lifecycleRestoreStorageClass": "Source storage class for restore"
            }
         }
      }
   ]
}
```

to the structure of messages coming from [EventBridge](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html): 
```json
{
  "version": "0",
  "id": "17793124-05d4-b198-2fde-7ededc63b103",
  "detail-type": "Object Created",
  "source": "aws.s3",
  "account": "111122223333",
  "time": "2021-11-12T00:00:00Z",
  "region": "ca-central-1",
  "resources": [
    "arn:aws:s3:::DOC-EXAMPLE-BUCKET1"
  ],
  "detail": {
    "version": "0",
    "bucket": {
      "name": "DOC-EXAMPLE-BUCKET1"
    },
    "object": {
      "key": "example-key",
      "size": 5,
      "etag": "b1946ac92492d2347c6235b4d2611184",
      "version-id": "IYV3p45BT0ac8hjHg1houSdS1a.Mro8e",
      "sequencer": "617f08299329d189"
    },
    "request-id": "N4N7GDK58NMKJ12R",
    "requester": "123456789012",
    "source-ip-address": "1.2.3.4",
    "reason": "PutObject"
  }
}  
```

Note, the data represented by these is the same, so it should just take a combination of a few serde `rename`, `alias`, and `flatten` attributes to get this to work for both types of events.

mmalenic:
Closes #119

### Changes
* Allow deserializing EventBridge messages as well as SQS messages.
* Add deserialization tests.
* Simplify `From` and `TryFrom` trait implementations.


brainstorm (approved):
Nice serde aliasing/renaming to have both event types (SQS and EventBridge) into one, kudos!

PR #124, opened 2024-02-26
fix(filemanager): re-run sqlx prepare with all targets

### LINKED ISSUE #123: filemanager: offline build does not work for all targets
When building with `--all-targets` the offline `.sqlx` queries do not allow compilation. `cargo sqlx prepare` should be run with `--all-targets` to account for this.

mmalenic:
Closes #123

### DIFF
--- a/lib/workload/stateless/filemanager/.sqlx/query-3fe88592da137c9706a5b541ef8edb7a794c72051d12c8c3b57d853b37a8392e.json
+++ b/lib/workload/stateless/filemanager/.sqlx/query-3fe88592da137c9706a5b541ef8edb7a794c72051d12c8c3b57d853b37a8392e.json
@@ -1,0 +1,20 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "select exists (select from information_schema.tables where table_name = 'object')",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "exists",
+        "type_info": "Bool"
+      }
+    ],
+    "parameters": {
+      "Left": []
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "3fe88592da137c9706a5b541ef8edb7a794c72051d12c8c3b57d853b37a8392e"
+}
--- a/lib/workload/stateless/filemanager/.sqlx/query-63da484d18ed2571639cfddadf992fe9ea92767a8274859de371f4e954e554b8.json
+++ b/lib/workload/stateless/filemanager/.sqlx/query-63da484d18ed2571639cfddadf992fe9ea92767a8274859de371f4e954e554b8.json
@@ -1,0 +1,116 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "select s3_object_id as \"s3_object_id!\",\n                object_id as \"object_id!\",\n                bucket,\n                key,\n                created_date,\n                deleted_date,\n                last_modified_date,\n                e_tag,\n                storage_class as \"storage_class: StorageClass\",\n                version_id,\n                created_sequencer,\n                deleted_sequencer,\n                number_reordered,\n                number_duplicate_events from s3_object",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "s3_object_id!",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 1,
+        "name": "object_id!",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 2,
+        "name": "bucket",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 3,
+        "name": "key",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 4,
+        "name": "created_date",
+        "type_info": "Timestamptz"
+      },
+      {
+        "ordinal": 5,
+        "name": "deleted_date",
+        "type_info": "Timestamptz"
+      },
+      {
+        "ordinal": 6,
+        "name": "last_modified_date",
+        "type_info": "Timestamptz"
+      },
+      {
+        "ordinal": 7,
+        "name": "e_tag",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 8,
+        "name": "storage_class: StorageClass",
+        "type_info": {
+          "Custom": {
+            "name": "storage_class",
+            "kind": {
+              "Enum": [
+                "DeepArchive",
+                "Glacier",
+                "GlacierIr",
+                "IntelligentTiering",
+                "OnezoneIa",
+                "Outposts",
+                "ReducedRedundancy",
+                "Snow",
+                "Standard",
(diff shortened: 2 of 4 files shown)


lib/workload/stateless/filemanager/.sqlx/query-3fe88592da137c9706a5b541ef8edb7a794c72051d12c8c3b57d853b37a8392e.json
```diff
@@ -0,0 +1,20 @@
+{
```
brainstorm, line 1: Should this be really committed in?

lib/workload/stateless/filemanager/.sqlx/query-63da484d18ed2571639cfddadf992fe9ea92767a8274859de371f4e954e554b8.json
```diff
@@ -0,0 +1,116 @@
+{
```
brainstorm, line 1: And this?

brainstorm (changes requested):
The `.sqlx` json files shouldn't be in-tree, or should they?

lib/workload/stateless/filemanager/.sqlx/query-63da484d18ed2571639cfddadf992fe9ea92767a8274859de371f4e954e554b8.json
mmalenic: Yes, I think this probably should be checked in. This is because it makes it much more convenient for anyone building Orcabus, e.g. via `cdk synth` or `cdk deploy`. It means that the CDK code doesn't need to be aware of filemanager's local database setup to work, as it allows compiling the crates without a docker database instance running. It is also mentioned as advice in the sqlx [docs](https://github.com/launchbadge/sqlx/blob/main/sqlx-cli/README.md#enable-building-in-offline-mode-with-query) (e.g. "Check this directory into version control").

mmalenic: Is there some disadvantage you can think of with regards to checking it in?

brainstorm: It just looks a bit odd because I expect code or uuids to be autogenerated by some reproducible process. For instance, imagine that we committed "cargo expand" outputs instead of just running it when we need to the debug some macro?

mmalenic: Yeah I get that, checking in auto generated code seems weird. I know this is different, but we check in lock files which are auto generated. In this case, instead of ensuring that package dependencies/versions are reproducible, storing the query files simplifies building the package. Another disadvantage of checking in the query files is that now we always have to keep them up to date when changing filemanager, which could be annoying.

In terms of alternatives, if there are no `query_*` files, any build of OrcaBus will need to start up the filemanager's docker postgres instance. To do this, we could potentially use a top-level `compose.yml`, or use the Makefiles? Not sure what the best approach is. Codebuild will also need to be aware of the postgres instance (although it needs to be aware of Rust too, which I think it currently isn't).

Maybe the best approach would be containerizing the whole filemanager build, which could simplify things for codebuild?

mmalenic: Okay fair enough, we might not have to dockerise, a top-level file could be enough. When I say top-level, I mean in the root of the repository. I think it's required here because the whole OrcaBus project needs to use `cdk synth` in some way, so the CDK code itself would need to be aware of filemanager's docker requirements.

/cc @victorskl @reisingerf as well, do you guys have any thoughts about this?

brainstorm: > Another tip on such DX dilemma is; keep an eye on the "upstream" toolchain activity e.g. the momentum around Cargo Lambda & its CDK support; vs the office AWS Rust SDK & its Lambda runtime move on this.

Upstream tooling such as `cargo-lambda` seems to focus on native deployment: https://www.cargo-lambda.info/

It even reads so in the subtitle of the tool:

> Run, Build, and Deploy Rust functions on AWS Lambda natively from your computer, **no containers or VMs required**.

brainstorm: I've seen that CDK construct before, btw @mmalenic: https://github.com/cargo-lambda/cargo-lambda-cdk

victorskl: Yup. Lambda docker runtime should be fine, I reckon. AWS seems to optimising it well. If the choice to docker runtime as Lambda target helps simplify things a bit; I'd say, let's go for it. I recall Will did the similar; and made a swap from Nodejs to container runtime - [ditto](https://github.com/umccr/orcabus/tree/346780c/lib/workload/stateless/metadata_manager/deploy/construct/app)

mmalenic: > What I got so far is roughly:
The issue is because of a few auto-generated files being checked in or not.
The recommendation was to check them in to avoid the dependency on a local/running DB. And there's an easy way to check whether they match the expected schema. A test which could be run automatically on commit/build to ensure those files are valid (effectively removing them from any review burden).
When they are present they make life a lot easier and if they are not the current way of building/deploying does not work anymore.

Yes, that's correct, except I'd also add:

- I don't think the pipeline build works yet anyway because there is no Rust and `cargo lambda` installed. Although, this issue could be dealt with separately to the current one.

The way this works is that filemanager's SQL dependency checks any queries against a real database at compile time, and emits errors produced by the database if the queries do not work. The `query_*` files are a sort of pre-compilation step that allows the dependency to recognise valid queries without a real database.

One thing I didn't mention, is that this feature can be turned off entirely (I.e. no real database checks at all). Although, I think it's a nice feature to have for dev.

brainstorm: Compile-time checks from sqlx are great, it saves a lot of time to see wiggle errors on your editor while editing the Rust/SQL logic without even having to deploy.

brainstorm (approved):

PR #125, opened 2024-02-27
fix(filemanager): correctly deal with null version_id

### LINKED ISSUE #122: filemanager: SQL queries do not consider null `version_id` values
When the SQL queries execute, they do not consider null `version_id` values as equal. Postgres yields `null` whenever the ordinary comparison operators are used (e.g. `=`, `<>`) and does not yield true/false if comparing null values (e.g. see [here](https://www.postgresql.org/docs/current/functions-comparison.html)). This means that objects without a `version_id` are not ingested correctly.

This can be fixed by using `is not distinct from` instead, which behaves correctly when comparing null values.

mmalenic:
Closes #122 

### Changes
* Correctly deals with null values in `version_id` when ingesting, reordering and detecting duplicates. Got to be careful of null comparisons in postgres :smiley_cat:.
* New tests for this case.


brainstorm (approved):
Ouch, good catch and fix!

PR #126, opened 2024-02-27
postgres-manager: Initiate postgres-manager

### LINKED ISSUE #99: new database per microservice
A mechanism to create database names for each microsrvice and remove passing master secret to stateless stack

Ref: https://github.com/umccr/orcabus/pull/92#discussion_r1479241166

williamputraintan:
closes #99 

This postgres manager will be responsible for 
1. new database creation for each microservice
2. create role for each microservice to use (could create rds-iam or conventional username-password role)
3. assign db owner to appropriate role

See [README](https://github.com/umccr/orcabus/blob/feature/microservice-database/lib/workload/stateless/postgres_manager/README.md) for lambda details.

(these lambda should be invoked manually)


lib/workload/stateless/postgres_manager/construct/postgresManager.ts
```diff
@@ -0,0 +1,127 @@
...
+      '/orcabus/db-cluster-resource-id'
+    );
+
+    // Let lambda access secret manager via aws managed lambda extension
+    // Ref:
+    // https://aws.amazon.com/blogs/compute/using-the-aws-parameter-and-secrets-lambda-extension-to-cache-parameters-and-secrets/
+    const lambdaLayerGetSecretExtension = lambda.LayerVersion.fromLayerVersionArn(
+      this,
+      'GetSecretExtensionLayer',
+      'arn:aws:lambda:ap-southeast-2:665172237481:layer:AWS-Parameters-and-Secrets-Lambda-Extension-Arm64:11'
```
andrewpatto, line 47: this feels brittle - I mean I realise there isn't much chance we want to spin up the stack in another region - but still is a bit of "magic string".

Of all the lambdas we make - aren't these ones going to be executed like 10 times total a month? Is it really worth optimising the cacheing of the secrets?

williamputraintan: Yeah fair enough, could make this the normal sdk secret fetch then

lib/workload/stateless/postgres_manager/function/create-pg-login-role.ts
```diff
@@ -0,0 +1,89 @@
...
+    throw new Error('this microservice is not configured for username-password authentication');
+  }
+
+  const client = new Client(pgMasterConfig);
+  await client.connect();
+  console.info('connected to RDS with master credential');
+
+  // create a new role
+  console.info('creating a user-password login role');
+  const password = pass_generator({ length: 10, numbers: true });
```
victorskl, line 36: Could you increase the length to `32`, pls

lib/workload/stateless/postgres_manager/function/create-pg-login-role.ts
```diff
@@ -0,0 +1,89 @@
...
+  const createRoleQueryTemplate = `CREATE ROLE ${microserviceName} with LOGIN ENCRYPTED PASSWORD '${password}'`;
+  await executeSqlWithLog(client, createRoleQueryTemplate);
```
victorskl, line 38: Could you change this to execute without logging, pls. Perhaps just `executeSql()` or something like that.

lib/workload/stateless/postgres_manager/function/create-pg-login-role.ts
victorskl: Perhaps, SM `GetRandomPasswordCommand` might be alternate here too. That'd streamline a bit without depending external lib.

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/secrets-manager/command/GetRandomPasswordCommand/

lib/workload/stateless/postgres_manager/function/create-pg-login-role.ts
```diff
@@ -0,0 +1,89 @@
...
+    Name: `orcabus/microservice/${microserviceName}`,
+    Description: `orcabus microservice secret for '${microserviceName}'`,
+    SecretString: JSON.stringify(secretValue),
+    Tags: [
+      { Key: 'stack', Value: 'manual' },
+      { Key: 'useCase', Value: 'store credential for the orcabus microservice' },
+      { Key: 'creator', Value: 'postgres_microservice at the orcabus stateless stack' },
+    ],
+  };
+  const smCommand = new CreateSecretCommand(smInput);
```
victorskl, line 62: While at it, would you mind set up rotation as well pls. That'd make the set up perfect. 

But, if you determine, you'd need more time working on rotation part then pls feel free to tackle as a new issue.

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/secrets-manager/command/RotateSecretCommand/

victorskl: And, we should rotate the master too..!

lib/workload/orcabus-stateless-stack.ts
```diff
@@ -73,6 +79,14 @@ export class OrcaBusStatelessStack extends cdk.Stack {
     //   However, the implementation is still incomplete...
   }
 
+  private createPostgresManager(config: PostgresManagerConfig) {
+    new PostgresManager(this, 'PostgresManager', {
+      ...config,
+      vpc: this.vpc,
+      lambdaSecurityGroup: this.lambdaSecurityGroup,
+    });
+  }
```
victorskl, line 88: Nice!

victorskl (approved):
Nice work, Will. Minor comments to attend.

lib/workload/stateless/postgres_manager/construct/postgresManager.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    const masterSecret = secretsmanager.Secret.fromSecretNameV2(
+      this,
+      'RdsMasterSecret',
+      props.masterSecretName
+    );
+
+    const dbClusterResourceId = ssm.StringParameter.valueFromLookup(
+      this,
+      '/orcabus/db-cluster-resource-id'
```
mmalenic, line 38: Could reuse the constant `dbClusterResourceIdParameterName` from `constants.ts` here?

mmalenic (approved):

lib/workload/stateless/postgres_manager/construct/postgresManager.ts
```diff
@@ -0,0 +1,117 @@
...
+      securityGroups: [props.lambdaSecurityGroup],
+    };
+
+    // 1. lambda responsible on db creation
+    const createPgDb = new nodejs.NodejsFunction(this, 'CreateDbPostgresLambda', {
+      ...rdsLambdaProps,
+      entry: __dirname + '/../function/create-pg-db.ts',
+      functionName: 'orcabus-create-pg-db',
+    });
+    masterSecret.grantRead(createPgDb);
```
mmalenic, line 64: I'm wondering if we could use this kind of pattern for database migration scripts too? I don't know if that would work though, because EdgeDb has it's own migration logic. Although, for anything using postgres directly, we could agree on some common database migration logic and re-use a Lambda function like this.

lib/workload/stateless/postgres_manager/construct/postgresManager.ts
williamputraintan: Yes, thanks! miss this one out

lib/workload/stateless/postgres_manager/function/create-pg-login-role.ts
williamputraintan: Yes thanks for feedback, I think would be better be part on a seperate PR along with the master secret rotation

#127

lib/workload/stateless/postgres_manager/construct/postgresManager.ts
williamputraintan: Yes, I think it make sense just that it would probably use its own microservice login credential and perhaps the migration lambda would live under its own microservice repo/folder?

reisingerf: Wouldn't you want to keep migrations (and everything that related to application specifics) together with / as part of the application?

mmalenic: Yeah that's true. I guess I was just thinking in terms of generic logic that applies SQL migrations irrespective of application logic. I.e. receives as input a bunch of `.sql` files and applies the migration, recording it in a migration table.

For example, sqlx [creates](https://github.com/launchbadge/sqlx/blob/27a49914ad5e0f21b8fb1f2a62db8d838724548f/sqlx-postgres/src/migrate.rs#L117-L129) a `_sqlx_migrations` table where it has the migration version, checksum, description, success status, etc.

reisingerf: Not sure if that's always a common approach across apps, but also you'd split your application concerns across multiple stacks / actors.

mmalenic: Yeah that's a good point. Probably it's cleanest to leave database table creation to the actual app that is using those tables.

PR #128, opened 2024-02-29
fix(postgres-manager): failing unit test from `cdk-nag`

williamputraintan:
- fix failing unit-test for postgres manager
	- add cdk-nag supression
	- ignore `tsc` check on microservice
- promote postgres-manager microservice from `cdk.construct` to `cdk.stack`

### DIFF
--- a/lib/workload/stateless/postgres_manager/deploy/postgres-manager-stack.ts
+++ b/lib/workload/stateless/postgres_manager/deploy/postgres-manager-stack.ts
@@ -1,4 +1,4 @@
-import { Duration } from 'aws-cdk-lib';
+import { Duration, Stack, StackProps } from 'aws-cdk-lib';
 import { Construct } from 'constructs';
 import * as lambda from 'aws-cdk-lib/aws-lambda';
 import * as nodejs from 'aws-cdk-lib/aws-lambda-nodejs';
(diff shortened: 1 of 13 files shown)


lib/workload/stateless/postgres_manager/deploy/postgres-manager-stack.ts
```diff
@@ -16,13 +16,13 @@ export type PostgresManagerConfig = {
...
 
-export type PostgresManagerStackProps = PostgresManagerConfig & {
+export type PostgresManagerProps = PostgresManagerConfig & {
   vpc: ec2.IVpc;
   lambdaSecurityGroup: ec2.ISecurityGroup;
 };
 
-export class PostgresManager extends Construct {
-  constructor(scope: Construct, id: string, props: PostgresManagerStackProps) {
+export class PostgresManagerStack extends Stack {
```
williamputraintan, line 24: @mmalenic am thinking of making microservice to stack instead of construct. One of the thought is that we could have the control for each microservice stack (e.g. we could deploy `yarn cdk-stateless-pipeline deploy ${stateless-microservice-stack}`  instead of the whole stateless stack). Do you have any thoughts or perhaps objection on this?

brainstorm (approved):

lib/workload/stateless/postgres_manager/deploy/postgres-manager-stack.ts
mmalenic: No objections, I think that's a good idea. What would be the process of deploying individual microservice stacks? Would there be a `individual_stacks.ts` file in `bin` that lists all the microservices, or is there a way deploy stack that are nested within other stacks?

lib/workload/stateless/postgres_manager/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "lambda-with-rds",
+  "name": "postgres-manager",
   "packageManager": "yarn@3.5.1",
   "dependencies": {
     "@aws-sdk/client-secrets-manager": "^3.515.0",
```
mmalenic, line 5: Should this be in the top-level `package.json`?

mmalenic (approved):

lib/workload/stateless/postgres_manager/deploy/postgres-manager-stack.ts
williamputraintan: Hmmm, perhaps adding in the stack construct at the `bin/orcabus.ts` could be it and `yarn cdk-orcabus ls`  could list all the stacks? I usually deploy from `yarn cdk-stateless-pipeline deploy ${stateless-microservice-stack}` as there might be some namespace clashes (if deployed the same stack in the same account). Not sure what is the best pattern/approach yet thou.

PR #132, opened 2024-03-04
Fix: Codebuild UnitTest

williamputraintan:
general:
- Commenting out staging and prod environment in stateless stack - stateful resource does not exist (in `stg` and `prod`) and the stateless lookup will fail the pipeline

metadata_manager:
- mark external dependencies for lambda runtime dependencies (to prevent unresolved path when bundling)

filemanager: 
- Rust installation for codebuild unit test  && `cargo-lambda`
- ~disable cdk-nag for filemanager due to `rust.aws-cdk-lambda` do not have `.toString()`  function which causes an error~ - it was due to compilation error led to the lambda construct to be undefined but is fixed on subsequent commit/pr


mmalenic:
Related to https://github.com/umccr/orcabus/pull/124. I think the Rust compilation is failing because of the `--all-targets --all-features` which compiles the tests in the Makefile.

victorskl (approved):

PR #136, opened 2024-03-06
fix(filemanager): ingest constraints

### LINKED ISSUE #131: filemanager: assuming `Object Removed` events occur when an object is overwritten
The current filemanager implementation assumes that an `Object Removed` event occurs when an object is replaced. However, this is not necessarily the case, and an object can be replaced using `PutObject` without triggering a `Removed` event. This causes a "duplicate key value violates unique constraint" error when trying to ingest the replaced object.

mmalenic:
Closes #131 

### Changes
* Make `version_id` required with a default value as the 'null' string.
* Ensure that only one row is returned from the main `current_objects` select query when reordering events.
* Various test fixes:
    * Add tests for missing created/deleted events, missing sequencer values, and multiple matching rows for reordering.
    * Fix permutation tests and add smaller non-ignored tests.

### DIFF
--- a/lib/workload/stateless/filemanager/database/migrations/0002_add_s3_object_table.sql
+++ b/lib/workload/stateless/filemanager/database/migrations/0002_add_s3_object_table.sql
@@ -24,6 +24,8 @@ create table s3_object (
     bucket text not null,
     -- The key of the object.
     key text not null,
+    -- The version id of the object.
+    version_id text not null default 'null',
     -- When this object was created. A null value here means that a deleted event has occurred before a created event.
     created_date timestamptz default null,
     -- When this object was deleted, a null value means that the object has not yet been deleted.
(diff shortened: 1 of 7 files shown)


lib/workload/stateless/filemanager/database/migrations/0002_add_s3_object_table.sql
```diff
@@ -24,6 +24,9 @@ create table s3_object (
     bucket text not null,
     -- The key of the object.
     key text not null,
+    -- The version id of the object. A 'null' string is used to indicate no version id. This matches logic in AWS which
+    -- also returns 'null' strings.
+    version_id text not null default 'null',
```
mmalenic, line 29: I know this looks weird, but I think that it is okay because AWS specifies that when the `version_id` is the string `'null'`, then versioning is not enabled: https://docs.aws.amazon.com/AmazonS3/latest/userguide/versioning-workflows.html. I.e. effectively, versioning is always enabled, and AWS uses `'null'` when it's disabled.

There is an alternative, and that's to have another constraint, .e.g. `created_sequencer_unique`, `deleted_sequencer_unique` and `versioning_id_unique`, and then adding adding additional insert queries that target the `versioning_id_unique` for de-duplication, although I feel that this adds a lot more repetitive code.

brainstorm: First option LGTM, weird but it's documented... I'd probably add that docs url in the comment itself for reference.

brainstorm (approved):

Issue #133, opened 2024-03-04
filemanager: consider using cargo-lambda-cdk instead of aws-cdk-lambda


mmalenic:
[This](https://github.com/cargo-lambda/cargo-lambda-cdk/tree/main) project is more up to date and maintained than the current [aws-cdk-lambda](https://github.com/rnag/rust.aws-cdk-lambda/tree/main) that filemanager uses. It also seems to have slightly simpler configuration, so we should investigate whether it's worthwhile to switch.

mmalenic:
[cargo-lambda-cdk](https://github.com/cargo-lambda/cargo-lambda-cdk/tree/main) also supports bundling the Lambda code using docker, which is convenient for CodeBuild because there is no requirement to install Rust or `cargo lambda`.

mmalenic:
@brainstorm might be able to remove those `query_*` files by using the before command hook to run the local database when building with cargo-lambda-cdk: https://github.com/cargo-lambda/cargo-lambda-cdk?tab=readme-ov-file#command-hooks

brainstorm:
> @brainstorm might be able to remove those `query_*` files by using the before command hook to run the local database when building with cargo-lambda-cdk: https://github.com/cargo-lambda/cargo-lambda-cdk?tab=readme-ov-file#command-hooks

Ohhh, yesyesyes!

PR #142, opened 2024-03-08
fix(filemanager): query files

mmalenic:
### Changes
* Updates query files from #136 which should fix the pipeline build.

Sorry @brainstorm, I think I'll be able to get rid of these query files soon with #133.


brainstorm (approved):
Indeed, garbage collecting those for good cannot come soon enough ;P

mmalenic:
Where is the garbage man in this city!

PR #143, opened 2024-03-10
Added feature icav2 copy batch state machine

alexiswl:
## Summary

This ICAv2 Copy Batch Utility wraps an AWS Step Function over the ICAv2 CopyBatch API.
This api is designed to copy a list of illumina file ids into a directory.

We exploit this API by taking in a manifest (a list of key values of source ids with their respective destinations) ( a source may have multiple destinations if it needs to be copied into a few different places ), and then monitor a set of API jobs to completion.

These CopyBatch API jobs have about a 20% fail rate, so we sometimes need to resubmit, this is built into the step function.

A 20% fail rate seems quite high, but when they work, these jobs can transfer 80 Gb of data in under 10 seconds so it's worth persisting with.

When working with the ICAv2 CopyBatch API, we need to provide a unique identifier for the run, and a unique location for the outputs.

Once all jobs are deployed we monitor their status (redeploy if failed - up to five times) and wait for all jobs to complete.

This process will transfer an entire BCLConvert process from one ICAv2 project to another in under 10 minutes!

Note that a current limitation prevents using this API within the same project.

## TODO List

- [x] Harmonise cdk deployment with the rest of the orcabus cdk deployment - versions, yaml.locks etc.  
- [x] Discuss whether this needs its own stateful component
- [x] Refactor so that each job runs in its own step function


victorskl (approved):
LGTM.! Yup, just needa hook up with outer CDK code would be good.

reisingerf (approved):

lib/workload/stateless/icav2_copy_batch_utility/step_functions_templates/copy_batch_state_machine.json
```diff
@@ -0,0 +1,188 @@
+{
+  "Comment": "Deploy a set of ICAv2 copy batch jobs and wait until theyre finished",
```
victorskl, line 2: Btw Alexis @alexiswl, I'd recommend to name these ASL into the naming convention `copy_batch_state_machine.asl.json`

There is benefit in IDE that support ASL syntax and, able to render step diagram. 

cc: @reisingerf

alexiswl: Oh no way! I was wondering if there was a JSON schema for this, thanks @victorskl, I'll do that now

victorskl: Bad news is - only avail in VSCode and draw.io at the mo. 😢 

Watch this space
https://github.com/aws/aws-toolkit-jetbrains/issues/584

victorskl:
Put this branch into draft mode until we finalise

alexiswl:
@victorskl this one is now ready for review

victorskl:
thanks Alexis for update, will review today.

lib/workload/stateless/stacks/icav2-copy-batch-utility/lambdas/check_or_launch_job/handler.py
```diff
@@ -0,0 +1,324 @@
...
+    "RUNNING"
+]
+
+
+# Try a job 10 times before giving up
+MAX_JOB_ATTEMPT_COUNTER = 10
+DEFAULT_WAIT_TIME_SECONDS = 10
+
+
+def handler(event, context):
```
victorskl, line 59: Non blocker comment (due to time constraint). 

One fine day down the track, we should unit test this handler to guardrail couple of `if` branching statements down under. Need to make sure these branching logics maintain throughout the system SDLC and, retain it when refactoring happens.

victorskl (approved):
LGTM from me.

PR #149, opened 2024-03-12
Postgres Manager Enhancement

williamputraintan:
- Modify the credential name for microservice secret 
	from `orcabus/microservice/${microserviceName}` to `orcabus/${microserviceName}/RdsLoginCredential` - i think the name makes more sense to what it is used for, as each microservice might have its own secret


williamputraintan:
currently deployed in dev @pdiakumis @victorskl @reisingerf @mmalenic

victorskl:
Yup, this makes sense, Will. 

Pls remember to create a GH ticket (or reuse existing one as see fit, pls)

victorskl:
Related #150

PR #152, opened 2024-03-14
refactor: swap rust function

### LINKED ISSUE #133: filemanager: consider using cargo-lambda-cdk instead of aws-cdk-lambda

mmalenic:
Closes #133. I think this fixes a few issues around building Rust code.

### Changes
* Swap the [aws-cdk-lambda](https://github.com/rnag/rust.aws-cdk-lambda/tree/main) for [cargo-lambda-cdk](https://github.com/cargo-lambda/cargo-lambda-cdk/tree/main)
* Filemanager is now compiled inside Docker, so CodeBuild no longer requires installing Rust and cargo-lambda.
* Importantly, the filemanager Lambda function is still deployed natively using a cargo-lambda, only the compilation occurs inside Docker.
* No more auto-generated `query-*` files.

These changes should make it easier to build the Filemanager code without special dependencies. There is a small downside in that the docker container produces code which isn't owned by the current user, meaning that sometimes permissions errors can occur if trying to remove/clean this directory. Although, I've documented this.


package.json
```diff
@@ -29,10 +29,10 @@
     "@aws-cdk/aws-apigatewayv2-integrations-alpha": "^2.110.0-alpha.0",
     "@aws-cdk/aws-lambda-python-alpha": "2.110.0-alpha.0",
     "aws-cdk-lib": "^2.126.0",
+    "cargo-lambda-cdk": "^0.0.19",
```
mmalenic, line 32: I'm wondering if this should go in a `package.json` inside the filemanager directory? I noticed @williamputraintan did this for the postgres manager.

williamputraintan: But then the pipeline/ environment no longer have access or install to this dependency as part of the CDK? The postgres manager has a `package.json` but it is for the runtime running the lambda

williamputraintan (approved):

victorskl (approved):
Nicely done.! I just tested locally and it works well.

victorskl:
It builds twice for each function though - migrate and ingest. But I reckon that is fine. We can optimise those bits later.

mmalenic:
> It builds twice for each function though - migrate and ingest. But I reckon that is fine. We can optimise those bits later.
Hmm, good point, that might slow things down a bit as there's definitely overlapping code (i.e. `filemanager` crate). I'll look into this.

package.json
mmalenic: Okay, sounds good. One approach could be to have multiple yarn workspaces as well? Although I think for now that's probably unnecessary for one dependency.

victorskl: Yup, yup. Well spotted, Marko. We're gonna hit the `yarn workspaces` in tick, real soon. Let sit tight for the moment, pls.

victorskl:
No rush. Early days...

PR #154, opened 2024-03-14
fix(pipeline): install Rust for unit tests

### LINKED ISSUE #153: fix: pipeline does not have Rust installed
Rust is still needed to run unit tests inside the CodeBuild pipeline, even if cargo-lambda is not required. Related to #152.

mmalenic:
Fixes #153

### Changes
* Add Rust back to pipeline in order to run unit tests.

Potentially another way around this is to dockerize the unit tests? Although not sure if something like that is necessary.


victorskl (approved):
lgtm

PR #155, opened 2024-03-15
Implemented SequenceRunManager app deployment

### LINKED ISSUE #139: Deployment go ahead with sequence_run_manager service
**Context:**

- As a part early explore on microservice approach; we have implemented the `sequence_run_manager` service
- Details about the implementation phase are captured in Trello card
- https://trello.com/c/TQX8yJuc/1445-orcabus-v1-implement-sequencerunmanager-microservice-app

**Caveats & Forces:**

_.. that drives this story card_

- `#include` #135
- `#include` https://github.com/umccr-illumina/ica_v2/issues/170
- https://umccr.slack.com/archives/C03ABJTSN7J/p1709773764245039

**Perks:**

- This decouples from the Portal Automation (leave the current production processes alone - sequencers data upload switch, v2-SampleSheet adoption, etc.)
- We can shut it off when the service is no longer needed without effecting anything upstream/downstream

**Actions:**

- Resume the implementation and deploy this service

victorskl:
* Added Shared API Gateway as stateful resource
* Introduced param constants that go across stateful & stateless
* Rearranged Python dependencies into deps directory
* Removed no longer used constants
* Removed now deprecated API Gateway Alpha constructs
* Updated CDK nag validation accordingly

Resolves #139


victorskl:
Anybody else pls feel free look into and chime in, if any.

williamputraintan (approved):
LGTM

lib/workload/stateful/apigw/component.ts
```diff
@@ -0,0 +1,42 @@
+import { Construct } from 'constructs';
+import { aws_ssm, Duration } from 'aws-cdk-lib';
+import * as cognito from 'aws-cdk-lib/aws-cognito';
+import { HttpUserPoolAuthorizer } from 'aws-cdk-lib/aws-apigatewayv2-authorizers';
+import { SHARED_HTTP_API_ID } from '../../../../config/param';
+import { CorsHttpMethod, HttpApi } from 'aws-cdk-lib/aws-apigatewayv2';
+
+export class SharedApiGatewayConstruct extends Construct {
```
mmalenic, line 8: Is the idea to reuse this for all microservices? For some reason I thought the API gateway would be stateless?

mmalenic (approved):

lib/workload/stateful/apigw/component.ts
reisingerf: Yes, agreed. Maybe refactor it out as a "common" construct that can be reused in (stateless) stacks?

williamputraintan: Possibly to put the construct directly on the top level of the stateless stack instead?

https://github.com/umccr/orcabus/blob/feature/pipeline-chores/lib/workload/orcabus-stateless-stack.ts#L53-L60

victorskl: Sound good, folks. I will move it back into its own service for now. 

The common construct sound like the plan. But let do that a little bit down the track - until we get to the point where we can collect few use cases to generalise. Then, we shall refactor their commonality from the specificness.

reisingerf (approved):

lib/workload/stateful/apigw/component.ts
victorskl: Have moved the APIGW into the service boundary. Pls see, if any.

victorskl:
@brainstorm @williamputraintan Heads up. This PR removes the Alpha constructs importing (which is deprecated upstream) and, they are available now in the main cdk library. i.e. It can be 1-1 replace import:

from `@aws-cdk/aws-apigatewayv2-alpha` to `aws-cdk-lib/aws-apigatewayv2`

mmalenic (approved):

victorskl:
Merging...

PR #156, opened 2024-03-15
fix: pipeline error caused by filemanager containerized build

mmalenic:
Related to #154.

### Changes
* Reverts changes made in #152 to build filemanager outside of a docker container.

This removes the containerized filemanager build because the build container is unable to connect to host.docker.internal inside CodeBuild to check queries. I'll have to change the compose/docker networking settings in order to re-enable to containerized build. I've tested this is dev, so for now, this should hopefully allow the build to succeed :crossed_fingers:.


victorskl (approved):
Early days.! :) Lgtm

PR #180, opened 2024-03-22
refactor: filemanager stack

mmalenic:
Closes #169 

### Changes
* As described in #169 except using `RDS_IAM` as part of postgres manager. Note, currently this still uses the master RDS credentials. Both these points will be addressed in #138.
* Move `CdkResourceInvoke` to `components`.
* Add stg and prod filemanager deployments.

PR #184, opened 2024-03-25
fix(filemanager): use docker run to get unique containers

mmalenic:
Related to #180 and #183.

Not sure exactly why the CodeBuild docker error occurs, but I think it's something to do with running `docker compose` in parallel. E.g. try running `docker compose up & docker compose up` and the same error occurs.

Either way, to hopefully fix this for good, the changes are:
* Use `docker compose run` in the CDK code, which always creates a unique container name with a random suffix.
    * Local dev can still use `docker compose up`.
* Use address and port `0.0.0.0:0` to create an arbitrary host:port mapping to the host.
* Rearrange `Makefile` and `.env.example` to reflect these changes.


victorskl (approved):
LGTM

Issue #169, opened 2024-03-18
filemanager: reconfigure deployment to match other microservices


mmalenic:
The deployment should be reconfigured to match other microservices. This includes:

* Change filemanager from construct to stack.
* Check cdk-nag tests to ensure filemanager passes/ignores relevant lints using the `microserviceStackArray` pattern.
* Make use of the postgres manager - related to #138.

mmalenic:
Using this to track a few other small changes:

* Change filemanager Lambda architecture from x86_64 to ARM64.
* Remove unused dependencies from filemanager crates

Issue #162, opened 2024-03-18
Add SSM session manager permission to CodeBuild role 


victorskl:
Change request: 

_DX, platform engineering_

Time to time, we will have to debug the CodeBuild. One technique is using `codebuild-breakpoint` and pause the build and, SSM enter into the build agent through session manager. Document outline below.
 
https://docs.aws.amazon.com/codebuild/latest/userguide/session-manager.html

For this to work, we need to attach the following permission to CodeBuild role. Perhaps, we should add this. So that no need to ad-hoc inline attaching it every other debugs...

```
{
  "Effect": "Allow",
  "Action": [
    "ssmmessages:CreateControlChannel",
    "ssmmessages:CreateDataChannel",
    "ssmmessages:OpenControlChannel",
    "ssmmessages:OpenDataChannel"
  ],
  "Resource": "*"
}
```

See https://umccr.slack.com/archives/C03ABJTSN7J/p1710754707635919 for an example debug run.

victorskl:
Alternatively we can do local debug.
https://docs.aws.amazon.com/codebuild/latest/userguide/use-codebuild-agent.html

But that "tax" us a bit to our local disk space.

```
docker images
public.ecr.aws/codebuild/amazonlinux2-aarch64-standard                 3.0       4a8e7923e567   13 days ago      9.74GB
```

victorskl:
Closing for now.

PR #187, opened 2024-03-25
fix(filemanager): silence Makefile output and trim lines

mmalenic:
Related to #180, #183, #184 and #169.

Hopefully this is the last of these, this time making sure by running tests using local CodeBuild.


victorskl (approved):
Getting there... 👍

victorskl:
> using local CodeBuild

Perhaps, check it out the `codebuild-breakpoint` technique that mentioned in #162 at some point in time.

PR #189, opened 2024-03-28
fix(filemanager): add uuid to target directory to avoid concurrency errors

mmalenic:
Related to #169 and associated PRs.

100% this should work now.

### Changes
* Add random uuid to target directory to prevent parallel execution of tests from interfering with another test's compiled outputs.
* Bump deps.

Thanks for the `codebuild-breakpoint` tip @victorskl, it helped a lot to find the issue.

### DIFF
--- a/lib/workload/stateless/filemanager/Cargo.lock
+++ b/lib/workload/stateless/filemanager/Cargo.lock
@@ -1542,9 +1542,9 @@ dependencies = [
 
 [[package]]
 name = "lambda_http"
-version = "0.10.0"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef8fafd7a4ce0bc6093cf1bed3dcdfc1239c27df1e79e3f2154f4d3299d4f60e"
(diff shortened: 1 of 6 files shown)


mmalenic:
A bit more context for this:

The issue was that because the cdk-nag tests run in parallel and the stacks are synthesized multiple times, another test would steal the compiled output from `cargo lambda` in the same `target` directory because it was expecting the exact same asset as all the other tests. This would result in a warning from `cargo lambda`:

```
WARN run: no binaries found in target directory after build, try using the --bin, --example, or --package options to build specific binaries
```

and subsequently cause the bundling to fail because aws-cdk-lib is unable to find any assets to bundle:

```
Bundling did not produce any output. Check that content is written to /tmp/cdk.outv5RzsD/bundling-temp-b3b38d7e2dc2fa1a2772516915a13b2482d6426c85b9386f68061db610f0b781
```

To fix this, the change was requiring that each invocation of `RustFunction` would compile output to a unique location. Note, that dependency compilation is in a shared cache so it's not that all code is recompiled for each test, just the final binary.

Interestingly, I wasn't able to trigger this very easily locally (although I did a few times), and it tended to only happen in CodeBuild.

Related #190.

victorskl (approved):
Great detective work!

PR #193, opened 2024-04-03
Feature/ica event pipe

reisingerf:
Added stateful construct to create:

- SQS queue for incoming ICA events
- DLQ for incoming queue
- CloudWatch Alarm to send notifications to Slack when events are moved to DLQ
- Event Pipe to forward ICA events to OrcaBus EventBus
- InputTransformation to remove SQS event wrapper when forwarding events to EventBus


config/constants.ts
```diff
@@ -49,6 +49,14 @@ const orcaBusStatefulConfig = {
     securityGroupName: lambdaSecurityGroupName,
     securityGroupDescription: 'allow within same SecurityGroup and rds SG',
   },
+  icaEventPipeProps: {
+    icaEventPipeName: 'orcabus-ica-event-pipe',
+    icaQueueName: 'orcabus-ica-event-queue',
+    icaQueueVizTimeout: 30,
+    eventBusName: eventBusName,
+    slackTopicArn: 'AwsChatBotTopic',
+    dlqMessageThreshold: 1,
```
victorskl, line 58: Perhaps, the DLQ threshold to the default 3?

victorskl (approved):
LGTM. Just minor comment.

config/constants.ts
victorskl: Perhaps, also to add a README too. 😁

reisingerf:
Updated version:
Code structure refactor and introduction of Stack in addition to Construct.

@williamputraintan please have a look a the changes to `OrcaBusStatefulDeploymentStage` which (I think) I had to modify to pass on the `env` with account and region. I think there are better ways, but that may be on your radar already.

lib/pipeline/orcabus-stateful-pipeline-stack.ts
```diff
@@ -148,6 +148,9 @@ class OrcaBusStatefulDeploymentStage extends cdk.Stage {
   ) {
     super(scope, environmentName, { env: { account: env?.account, region: 'ap-southeast-2' } });
 
-    new OrcaBusStatefulStack(this, 'OrcaBusStatefulStack', stackProps.orcaBusStatefulConfig);
+    new OrcaBusStatefulStack(this, 'OrcaBusStatefulStack', {
+      env: env,
+      ...stackProps.orcaBusStatefulConfig,
+    });
   }
```
williamputraintan, line 155: Perhaps we could just put new pipe stack here instead inside of the  stateful stack?

williamputraintan:
> modify to pass on the env with account and region. I think there are better ways, but that may be on your radar already.
I think it just more explicit, I suppose the default should already come from the stage environment.

lib/workload/stateful/ica_event_pipe/stack.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (!props.env) {
+      throw new Error('No env properties found. Please ensure env.account and evn.region are set.');
+    }
+    if (!props.env.account) {
+      throw new Error('No account');
+    }
+    if (!props.env.region) {
+      throw new Error('No region');
+    }
+    return 'arn:aws:sns:' + props.env.account + ':' + props.env.region + ':' + props.slackTopicName;
```
victorskl, line 52: Flo is struggling with bloody, slippery `env` here. 😄 
(All good though, seriously. No action need.)

lib/pipeline/orcabus-stateful-pipeline-stack.ts
victorskl: From my side, could I leave refactoring `OrcaBusStatefulStack` with you, Will. You are the CodePipeline expert - so, I wouldn't want to mess up the setup done there.

But, yah, Flo is having a little teething issue with missing alembic, the `env`. We both puzzle the other days.

victorskl (approved):
Looking good, too, Flo.!

williamputraintan (approved):

lib/pipeline/orcabus-stateful-pipeline-stack.ts
reisingerf: I could move my stack here, yes. 
However, as Victor said, I don't want to interfere with your plans. I thought the `OrcaBusStatefulStack` is supposed to become a wrapper class instead of a stack. 

Either way, happy to go with what works best. Here I just point out one of the "issues" I ran into.

reisingerf:
OK, I'll merge for now and we'll clean up the stack-in-stack thing once we have a better solution.

PR #194, opened 2024-04-04
Update Stateless Pipeline Test

williamputraintan:
It seems that cdk-ing the pipeline stack will also synth all the children stacks (microservice stack), but we also have a dedicated test for these children stacks. The `cdk-nag` test only works on the top layer stack and for that reason we have the deployment (children) stack to be tested separately.

The solution is to mock the children stack with a mock stack when testing the pipeline so it doesn't expose the real micro-stack. This should half the building time for testing.

Thanks @mmalenic for pointing the double build when testing!

### DIFF
--- a/test/stateless/stateless-pipeline.test.ts
+++ b/test/stateless/stateless-pipeline.test.ts
@@ -1,8 +1,18 @@
-import { App, Aspects } from 'aws-cdk-lib';
+import { App, Aspects, Stack } from 'aws-cdk-lib';
 import { Annotations, Match } from 'aws-cdk-lib/assertions';
 import { SynthesisMessage } from 'aws-cdk-lib/cx-api';
 import { AwsSolutionsChecks, NagSuppressions } from 'cdk-nag';
 import { StatelessPipelineStack } from '../../lib/pipeline/orcabus-stateless-pipeline-stack';
+
+// we are mocking the deployment stack here, as we have a dedicated cdk-nag test for deployment stack
+// see the ./stateless-deployment.test.ts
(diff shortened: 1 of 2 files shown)


test/stateless/stateless-pipeline.test.ts
```diff
@@ -1,37 +1,42 @@
...
+  NagSuppressions.addStackSuppressions(stack, [
+    { id: 'AwsSolutions-IAM4', reason: 'Allow CDK Pipeline' },
+    { id: 'AwsSolutions-IAM5', reason: 'Allow CDK Pipeline' },
+    { id: 'AwsSolutions-S1', reason: 'Allow CDK Pipeline' },
+    { id: 'AwsSolutions-KMS5', reason: 'Allow CDK Pipeline' },
+    { id: 'AwsSolutions-CB3', reason: 'Allow CDK Pipeline' },
+  ]);
+
   test('cdk-nag AwsSolutions Pack errors', () => {
     const errors = Annotations.fromStack(stack)
```
williamputraintan, line 41: @mmalenic I didn't look into detailed to the implementation of the `fromStack` that forces to multiple build on every call. But just from the build log, the build only happen once despite the `fromStack` function being called multiple times.

mmalenic: Nice! I guess there's some build caching going on?

The Rust code in filemanager is compiling multiple times anyway even with a normal `cdk synth` (i.e. outside the context of parallel test execution) - I'll look into why this is happening, and I don't think it's related to the tests.

mmalenic (approved):

victorskl (approved):
Good stuff, folks!

PR #196, opened 2024-04-05
feat(filemanager): iam rds credentials

### LINKED ISSUE #138: filemanager: link up filemanager with postgres database manager
The filemanager should utilize the postgres manager's database creation and credential logic to create a filemanager database. This involves linking up the filemanager deployment code with the postgres manager and generating/obtaining database credentials within the Lambda function. Ideally, `DbAuthType.RDS_IAM` should be used for IAM authentication.

It doesn't look like the Rust sdk supports a `generate_rds_iam_token` function which means tokens must generated manually. See https://github.com/awslabs/aws-sdk-rust/issues/951.

mmalenic:
Closes #138 

### Changes
* Generate RDS IAM credentials in order to connect to the filemanager database.
    * This uses `DbAuthType.RDS_IAM` postgres manager setting.
    
### Misc
* Add a name to the event source dead letter queue.
* Filemanager lambda functions now use a shared database pool created in the init lambda execution phase.

### DIFF
--- a/lib/workload/orcabus-stateless-stack.ts
+++ b/lib/workload/orcabus-stateless-stack.ts
@@ -12,7 +12,7 @@ import {
 } from './stateless/postgres_manager/deploy/postgres-manager-stack';
 import { SequenceRunManagerStack } from './stateless/sequence_run_manager/deploy/component';
 import { EventBus, IEventBus } from 'aws-cdk-lib/aws-events';
-import { StringParameter } from 'aws-cdk-lib/aws-ssm';
+import { ISecret, Secret } from 'aws-cdk-lib/aws-secretsmanager';
 
 export interface OrcaBusStatelessConfig {
   multiSchemaConstructProps: MultiSchemaConstructProps;
@@ -27,6 +27,7 @@ export class OrcaBusStatelessStack extends cdk.Stack {
   private readonly vpc: IVpc;
   private readonly lambdaSecurityGroup: ISecurityGroup;
   private readonly mainBus: IEventBus;
+  private readonly rdsMasterSecret: ISecret;
 
   // microservice stacks
   microserviceStackArray: cdk.Stack[] = [];
@@ -43,6 +44,11 @@ export class OrcaBusStatelessStack extends cdk.Stack {
       'OrcaBusLambdaSecurityGroup',
       props.lambdaSecurityGroupName,
       this.vpc
+    );
+    this.rdsMasterSecret = Secret.fromSecretNameV2(
+      this,
+      'FilemanagerDatabaseSecret',
+      props.rdsMasterSecretName
     );
 
     this.mainBus = EventBus.fromEventBusName(this, 'OrcaBusMain', props.eventBusName);
@@ -89,18 +95,11 @@ export class OrcaBusStatelessStack extends cdk.Stack {
       )
     );
 
-    const clusterIdentifier = StringParameter.valueFromLookup(
-      this,
-      config.databaseClusterIdentifierParameter
-    );
-
     return new Filemanager(this, 'Filemanager', {
       buckets: config.eventSourceBuckets,
       buildEnvironment: {},
-      host: clusterIdentifier,
+      host: this.rdsMasterSecret.secretValueFromJson('host').unsafeUnwrap(),
       port: config.port,
-      database: config.database,
-      user: config.user,
       securityGroup: this.lambdaSecurityGroup,
       eventSources: [queue],
       migrateDatabase: true,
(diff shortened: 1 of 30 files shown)


lib/workload/orcabus-stateless-stack.ts
```diff
@@ -88,17 +94,13 @@ export class OrcaBusStatelessStack extends cdk.Stack {
...
-      'FilemanagerDatabaseSecret',
-      config.databaseSecretName
-    );
 
     return new Filemanager(this, 'Filemanager', {
       buckets: config.eventSourceBuckets,
       buildEnvironment: {},
-      databaseSecret,
-      databaseSecurityGroup: this.lambdaSecurityGroup,
+      host: this.rdsMasterSecret.secretValueFromJson('host').unsafeUnwrap(),
```
mmalenic, line 101: As an alternative, the `host` could be defined in an SSM parameter, instead of using the master secret property.

lib/workload/stateful/event_source/component.ts
```diff
@@ -55,7 +55,10 @@ export class EventSource extends Construct {
   constructor(scope: Construct, id: string, props: EventSourceProps) {
     super(scope, id);
 
-    this.deadLetterQueue = new Queue(this, 'DeadLetterQueue', { enforceSSL: true });
+    this.deadLetterQueue = new Queue(this, 'DeadLetterQueue', {
+      queueName: 'orcabus-event-source-dlq',
```
brainstorm, line 59: Hardcoded queue name should probably go under constants/props?

lib/workload/stateless/filemanager/filemanager/src/env.rs
```diff
@@ -1,10 +1,10 @@
 //! Handle environment variables.
 //!
 
-use dotenvy::dotenv;
 use std::ffi::OsStr;
```
brainstorm, line 4: Interesting, why is ffi::OsStr needed here?

brainstorm (approved):
Two minor suggestions otherwise LGTM!

Also higher amount of explicit lifetimes `<'a>` in this changeset, what motivated the compiler to have them defined?

lib/workload/orcabus-stateless-stack.ts
williamputraintan: Agree! Perhaps we could have SSM to store the `port` and `host` on the stack where the db is created.

williamputraintan (approved):

lib/workload/stateful/event_source/component.ts
mmalenic: I thought `constants.ts` was getting a bit long already. I hard-coded it because there is no requirement in the current stack to set this as configurable, i.e. there is no dependency from the stateful stack to the stateless stack for the name of the dead letter queue.

Although probably for symmetry with the non-dead letter queue this could be configurable?

lib/workload/stateless/filemanager/filemanager/src/env.rs
mmalenic: It doesn't have to be, I was just copying the `env::var` types: https://doc.rust-lang.org/std/env/fn.var.html, although it could just as well be `K: AsRef<str>`

lib/workload/orcabus-stateless-stack.ts
mmalenic: The port is configurable from the top-level `constants.ts`, although annoyingly the endpoint url isn't! I can add some SSM parameters which probably make more sense than the master rds secret.

lib/workload/stateful/event_source/component.ts
victorskl: Perhaps, suffix `-dlq` to the corresponding main queue name. i.e. `props.queueName + '-dlq'` something like that...

victorskl (approved):

lib/workload/stateless/filemanager/filemanager/src/database/mod.rs
```diff
@@ -2,31 +2,83 @@
...
+    async fn generate_password(&self) -> Result<String>;
+}
+
 /// A database client handles database interaction.
 #[derive(Debug)]
-pub struct Client {
-    pool: PgPool,
+pub struct Client<'a> {
+    // Use a Cow here to allow an owned pool or a shared reference to a pool.
+    pool: Cow<'a, PgPool>,
```
mmalenic, line 27: > Also higher amount of explicit lifetimes <'a> in this changeset, what motivated the compiler to have them defined?

@brainstorm the compiler requires this because I'm using a `Cow` :cow: here which is sort of like a shared-reference smart pointer that can store an owned or reference type for the database pool. This means that everything that uses the database `Client` now has to have a lifetime defined. 

It would be cool if lifetimes could be elided in structs too but I don't think this is planned: https://github.com/rust-lang/rfcs/blob/master/text/0141-lifetime-elision.md#lifetime-elision-in-structs

lib/workload/stateful/event_source/component.ts
mmalenic: Yes, I think that's a good in between.

PR #211, opened 2024-04-09
Fixed Token Service initial value json parsing

victorskl:
* Removed static endpoint lookup requirement from
  boto client bootstrapping


victorskl:
@reisingerf Comment if any. Merging ahead to a try for now.

PR #157, opened 2024-03-15
Update Readme and Refactor Pipeline Code

williamputraintan:
- Shorten `yarn cdk` syntax from `yarn cdk-stateless-pipeline` -> `yarn cdk-stateless`, the pipeline should just be the stack 
- `make suite` -> `make test-suite` which only test microservice tests
- delete `bin/orcabus.ts` , alternative to deploy directly from local is from `yarn cdk-stateless deploy {theBetaStack}` 
- Updated docs (Happy for any feedback here!)
	- Update Readme.md
	- Update the MICROSERVICE.md 
	- Added SHARED_RESOURCE.md to describe what resources is being shared within the stateful stack 
	
Fix #150 

Could be merged after #155

PR #216, opened 2024-04-11
chore: Split Stateful Single Stack to Multi Stack

### LINKED ISSUE #188: Rename/Split Shared Resources in Stateful Deployment Stack
In the deployment of the stateful pipeline, all resources are currently in 1 stack. We should split or rename this to become a shared resource stack as there might be multiple service requiring its own stateful resource. Isolating the shared resource from other service resources could make easier deployment testing and breaking it wouldn't affect other services.

williamputraintan:
- mainly refactoring that shared resource are all in one stack and any other µ-app that require stateful be on different stack (preventing new µ-app feature to step on SharedStack's toe)
- wrapping multiple stateful stacks in a class called `StatefulStackCollection` when defined in stage
- refactoring on typescript naming convention

Example on `cdk ls` command after refactoring:

```bash
yarn cdk-stateful ls

OrcaBusStatefulPipeline
OrcaBusStatefulPipeline/BetaDeployment/IcaEventPipeStack (BetaDeployment-IcaEventPipeStack)
OrcaBusStatefulPipeline/BetaDeployment/SharedStack (BetaDeployment-SharedStack)
OrcaBusStatefulPipeline/BetaDeployment/TokenServiceStack (BetaDeployment-TokenServiceStack)
...
```

resolves #188


README.md
```diff
@@ -103,12 +103,25 @@ _Heads up: Polyglot programming environment. We shorten some trivial steps into
...
+Guide](https://docs.aws.amazon.com/prescriptive-guidance/latest/best-practices-cdk-typescript-iac/typescript-best-practices.html#naming-conventions).
+
+- Use camelCase for variable and function names.
+- Use PascalCase for class names and interface names.
+- Use camelCase for interface members.
+- Use PascalCase for type names and enum names.
+- Name files with camelCase (for example, ebsVolumes.tsx or storage.tsb)
+
+For folder name, we will be using `kebab-case`
+
```
williamputraintan, line 124: I have been inconsistent with naming convention, but I think we could follow what AWS suggestion here.

Folder naming convention is not mentioned, but in general `kebab-case` is more common. For example, AWS uses `kebab-case` in their CDK library (`import { IVpc } from 'aws-cdk-lib/aws-ec2';`)

williamputraintan: will refactor on the stateless stack after this PR along with removing the parent (of the µ-app) stack

https://github.com/umccr/orcabus/pull/157#issuecomment-2000650031

williamputraintan:
This might need a redeployment of resources here as stack id will change.

We could also roll this to all of our account env (dev, stg, prod) once this is approved

victorskl:
.. will review today

victorskl:
reviewing...

README.md
victorskl: Sound good. I support this.

docs/developer/SHARED_RESOURCES.md
```diff
@@ -1,44 +1,7 @@
 # Shared Resources
```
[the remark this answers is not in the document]
victorskl: Feel, 😬 a little bit of double efforts/handling here, yes. 

Up to the team with favour the change/remove; ie stick the README closer to the stack/resource that is deployed. At least for these early days taking off phases. We tend to change .. a lot, frequently.

lib/workload/components/vpc/index.ts
victorskl, line 1: We can also remove this VPC lookup function i.e. rather promote to use the `vpcProps` pass down from the `constants.ts` and, reconstruct at the stack level. Like the way it does in Token Service stack.

What chu think, Will? Leaving it is fine, too.

lib/workload/stateful/stacks/shared/constructs/database/index.ts
```diff
@@ -112,7 +112,7 @@ export class Database extends Construct {
   constructor(scope: Construct, id: string, props: DatabaseProps) {
     super(scope, id);
 
-    const dbSecret = new rds.DatabaseSecret(this, id + 'DbSecret', {
```
victorskl, line 115: Thank you for cleaning this id thingy up.! 😬

lib/workload/stateful/stacks/shared/constructs/schemaregistry/index.ts
victorskl, line 1: Bye, bye, my dear "component". Hehe. J/K 😬

lib/workload/stateful/stacks/token-service/deploy/stack.ts
```diff
@@ -35,7 +36,7 @@ export class TokenServiceStack extends Stack {
     super(scope, id, props);
     this.props = props;
 
-    this.vpc = Vpc.fromLookup(this, 'MainVpc', props.vpcProps);
+    this.vpc = getVpc(this);
```
victorskl, line 39: Ah. You want the function one. Please check with the Flo! 😃 Think, he prefer the other way, `vpcProps` pass down. Leave it with you two.

lib/workload/stateful/statefulStackCollectionClass.ts
```diff
@@ -0,0 +1,47 @@
...
+import { TokenServiceProps, TokenServiceStack } from './stacks/token-service/deploy/stack';
+import { IcaEventPipeStack, IcaEventPipeStackProps } from './stacks/ica-event-pipe/stack';
+
+export interface StatefulStackCollectionProps {
+  sharedStackProps: SharedStackProps;
+  tokenServiceStackProps: TokenServiceProps;
+  icaEventPipeStackProps: IcaEventPipeStackProps;
+}
+
+export class StatefulStackCollection {
```
victorskl, line 14: Oh, the "collection" class. I like this naming.! Good one, Will. 👍

victorskl (approved):
👍 from me.!

lib/workload/components/vpc/index.ts
williamputraintan: Agree on lookup at µ-stack level rather than passing `IVpc` as a stack props. I feel it is more isolated and it is simpler to get onboard when dev the µ-app rather depending on resource as the props.

williamputraintan: I just thought this function is just a quick single-line wrapper of common resource lookup

lib/workload/stateful/statefulStackCollectionClass.ts
williamputraintan: Initially was set to "manager" class, but I think we had enough managers in OrcaBus 🤔

lib/workload/stateful/stacks/token-service/deploy/stack.ts
reisingerf: Hm,... does the function add much? 
It's a bit easier for me without... first I don't need to know about any outside function and second it's just a lookup like I'd need to do for any other (provided) resource.
Also, `getVpc()` hides away configuration details that otherwise would be in the usual, expected spot (config constants), right?

reisingerf (approved):
Looks nice and neat! Go for it!

lib/workload/stateful/stacks/token-service/deploy/stack.ts
victorskl: Yup. Since we change to independent "stack" concept, the props password become more make sense.

Think, Will agree (somewhere ^ there comment; deprecating this function lookup all together). Perhaps, Will could help make necessary changes and, use `vpcProps` pass down. I reckon, that should work... But, leaving the final call to Will.

williamputraintan: Will remove it! It will be easier for native lookup rather than understanding a custom function.

victorskl: Yay! 🎉

lib/workload/stateful/statefulStackCollectionClass.ts
```diff
@@ -0,0 +1,51 @@
...
+      ...statefulConfiguration.tokenServiceStackProps,
+    });
+
+    this.icaEventPipeStack = new IcaEventPipeStack(scope, 'IcaEventPipeStack', {
+      ...this.createTemplateProps(env, 'IcaEventPipeStack'),
+      ...statefulConfiguration.icaEventPipeStackProps,
+    });
+  }
+
+  private createTemplateProps(env: Environment, serviceName: string): StackProps {
```
williamputraintan, line 41: added a few more initial tagging template here for each µ-stack @victorskl @reisingerf

reisingerf: Nice!

reisingerf (approved):

PR #218, opened 2024-04-12
fix(filemanager) change i32 to i64 to support bigger sizes

### LINKED ISSUE #217: filemanager: use i64 instead of i32
The data type on the filemanager's event [message](https://github.com/umccr/orcabus/blob/b0fc25849a4922a18ec407dc903baa29f7aa9e7e/lib/workload/stateless/filemanager/filemanager/src/events/aws/message.rs#L99) size should be i64 to accommodate larger file sizes.

mmalenic:
Closes #217

### Changes
* Replace i32 with i64.
* Replace integer with bigint.


lib/workload/stateless/filemanager/filemanager/src/database/aws/ingester.rs
```diff
@@ -291,6 +291,35 @@ pub(crate) mod tests {
...
+    #[sqlx::test(migrator = "MIGRATOR")]
+    async fn ingest_object_created_large_size(pool: PgPool) {
+        let mut events = test_events();
+        events.object_deleted = Default::default();
+
+        events
+            .object_created
+            .sizes
+            .iter_mut()
+            .for_each(|size| *size = Some(373588380934135341));
```
brainstorm, line 303: I'd substitute this magic number for some `std::` constant if possible?

lib/workload/stateless/filemanager/filemanager/src/database/aws/ingester.rs
```diff
@@ -291,6 +291,35 @@ pub(crate) mod tests {
...
+        let ingester = test_ingester(pool);
+        ingester.ingest_events(events).await.unwrap();
+
+        let (object_results, s3_object_results) = fetch_results(&ingester).await;
+
+        assert_eq!(object_results.len(), 1);
+        assert_eq!(s3_object_results.len(), 1);
+        assert_with(
+            &s3_object_results[0],
+            Some(373588380934135341),
```
brainstorm, line 314: Ditto

lib/workload/stateless/filemanager/filemanager/src/events/aws/message.rs
```diff
@@ -162,13 +162,58 @@ impl From<Message> for FlatS3EventMessages {
...
+                    "arn:aws:s3:::bucket"
+                ],
+                "detail": {{
+                    "version": "0",
+                    "bucket": {{
+                        "name": "bucket"
+                    }},
+                    "object": {{
+                        "key": "key",
+                        "size": 373588380934135341,
```
brainstorm, line 194: Ahh, I see where the number comes from now... maybe we can be cheeky and alter the example so that it matches an already known constant in std? ;)

brainstorm (changes requested):

lib/workload/stateless/filemanager/filemanager/src/events/aws/message.rs
mmalenic: Good call, I'll replace it with [`i64::MAX`](https://doc.rust-lang.org/std/primitive.i64.html#associatedconstant.MAX). May as well test the biggest value here.

brainstorm (approved):
Delicious! ;)

PR #227, opened 2024-04-16
Fixed Token Service initial value json parsing

victorskl:
* It is better to align Service User secret in JSON
  format container as initial value. Otherwise, JSON
  parsing error could mis-lead the admin user who'd
  reset this value as initial bootstrap step.


reisingerf (approved):

victorskl:
Merging...

PR #228, opened 2024-04-17
feat(filemanager): add external id and attribute column

### LINKED ISSUE #223: filemanager: allow storing arbitrary data for objects
There should be an ability to store some arbitrary data on an object to support uses cases by external services. In terms of implementation, this could be as simple as adding a `jsonb` field to the object/s3_object table.

I'm not sure if this best placed on [`object`](https://github.com/umccr/orcabus/blob/main/lib/workload/stateless/filemanager/database/migrations/0001_add_object_table.sql) or [`s3_object`](https://github.com/umccr/orcabus/blob/main/lib/workload/stateless/filemanager/database/migrations/0002_add_s3_object_table.sql) though. This is important when considering the same object which is present in multiple locations. For example, one `object` can contain multiple `s3_objects` (due to the [references `object_id`](https://github.com/umccr/orcabus/blob/3353946f680901a0ec4164f4705ca7c47adee007/lib/workload/stateless/filemanager/database/migrations/0002_add_s3_object_table.sql#L20) on the `s3_object` table), i.e. a single object can have multiple locations. Does this mean data about an object should be applied to the location (i.e. the `s3_object`), or the actual object (i.e. the `object` table). Maybe there could be `jsonb` fields on both tables?

### LINKED ISSUE #224: filemanager: expose an external identifier for objects
Each [`object`](https://github.com/umccr/orcabus/blob/3353946f680901a0ec4164f4705ca7c47adee007/lib/workload/stateless/filemanager/database/migrations/0001_add_object_table.sql#L4)/[`s3_object`](https://github.com/umccr/orcabus/blob/3353946f680901a0ec4164f4705ca7c47adee007/lib/workload/stateless/filemanager/database/migrations/0002_add_s3_object_table.sql#L18) in the filemanager database has a `uuid` primary key which never updates, however there should probably be another external id for clarity, that can be referred to and also never changes. This is important if merging `s3_objects` into one `object`. If the external id never changes then when an `s3_object` is linked to an `object`, there could be multiple external ids that refer to the same object. This isn't an issue as long as the database keeps track of this possibility and doesn't forgot old external ids.

mmalenic:
Closes #224, Closes #223

### Changes
* Add public-facing external id for filemanager objects. Note, I'm intentionally editing the .sql files without introducing new ones as long as possible in order to maintain readability.
* This id should returned/received by filemanager in queries.
* Add attribute columns which can store any json data.


mmalenic:
@brainstorm I've added an attribute column for #223 in this PR as well, because it's just a straight-forward extra column without any logic being affected. I've also made a note of the local database compilation/migration details.

brainstorm (approved):
Cool! Not only the external id bit (which I agree it can be handy), but also the semi-automated migration experiment you were discussing in slack, cool!

Issue #167, opened 2024-03-18
Refactor Config File


williamputraintan:
Configuration file is quite messy where everything is cramp into 1 file which includes all configuration including microservices and AWS infrastructure config. I think we could refactor this out so each µ-service/AWS resource has its own config file. 

The refactor should also allow the infrastructure deployment without relying on specific account number (that is currently the config is tied to). Perhaps adding this to docs would also be helpful for isolation testing.

Action items:

- [X] Use type checking from each stack when defining configuration
- [ ] A mechanism so config can be independent from account number #237

williamputraintan:
On independent deployment point, we could simply remove the account number and let use the default CDK env. But, then there are some resources that the OrcaBus assume that it exist (e.g. cognito, VPC) so perhaps we should have docs to deploy a prerequisite stack/IaC before trying to deploy OrcaBus Stacks.

williamputraintan:
Tracking this on different ticket https://github.com/umccr/orcabus/issues/237, and this issue is focus on refactoring config file to be readable.

PR #231, opened 2024-04-17
chore: splitting config file

williamputraintan:
Just the initial phase of refactoring config file as now seems to be a bit tedious. Splitting the config per individual stack, so could make use of the Type check for each stack props.

Partial #167


victorskl (approved):

reisingerf (approved):
NICE!

Issue #233, opened 2024-04-18
chore: CodePipeline Stage name refactor


victorskl:
**Context:**

From `BetaDeployment-TokenServiceStack` --> `OrcaBusBeta-TokenServiceStack`

This would entails another tear down and, re-init. But, we shall probably do this before we put it into prod. 

ETA -> end next week (end April) or, the following week (in May)

Wait, current ongoing testing to run pass first.

PR #250, opened 2024-04-24
event archive bucket policy modification according different env

raylrui:
Fix/Refactor of event archiver: 

1. RemovalPolicy:
`RemovalPolicy.DESTROY `mode in our dev/test envs, 
only enable `RemovalPolicy.RETAIN` in prod env. 
For dev/test redeployment, we may still delete all contents in the bucket and get the bucket physically deleted.

2. unique bucket name 
beta: orcabus-event-archive-_`devAccountId`_, 
gamma: orcabus-event-archive-_`stgAccountId`_, 
prod: orcabus-event-archive-_`prodAccountId`_, 

From discuss: https://umccr.slack.com/archives/C03ABJTSN7J/p1713927722151309

### DIFF
--- a/config/constants.ts
+++ b/config/constants.ts
@@ -4,6 +4,20 @@ export type AccountName = 'beta' | 'gamma' | 'prod';
 
 export const region = 'ap-southeast-2';
 
+/**
+ * accountIdAlias: mapping from AccountName to AWS Account ID.
+ * accountIdAliasType to ensure that the accountId is always explicitly defined for each account.
+ */
+type accountIdAliasType = {
+  [K in AccountName]: string;
+};
+
+export const accountIdAlias: accountIdAliasType = {
+  beta: '843407916570', // umccr_development
+  gamma: '455634345446', // umccr_staging
+  prod: '472057503814', // umccr_production
+};
+
 // external ICA constants
 export const icaAwsAccountNumber = '079623148045';
 
@@ -63,7 +77,7 @@ export const eventBusName = 'OrcaBusMain';
 export const eventSourceQueueName = 'orcabus-event-source-queue';
 
 // Event Archiver constants for EventBus Contruct in SharedStack
-export const archiveBucketNamePrefix = 'orcabus-event-archive-';
+export const archiveBucketNameSuffix = '-orcabus-event-archive';
 export const archiveSecurityGroupName = 'OrcaBusSharedEventBusEventArchiveSecurityGroup';
 
 /**
--- a/config/constants.ts
+++ b/config/constants.ts
@@ -63,8 +63,8 @@ export const eventBusName = 'OrcaBusMain';
 export const eventSourceQueueName = 'orcabus-event-source-queue';
 
 // Event Archiver constants for EventBus Contruct in SharedStack
-export const archiveBucketName = 'event-archive-bucket';
-export const archiveSecurityGroupName = 'OrcaBusEventArchiveSecurityGroup';
+export const archiveBucketNameSuffix = 'event-archive-bucket';
+export const archiveSecurityGroupName = 'OrcaBusShareEventBusEventArchiveSecurityGroup';
 
 /**
  * Configuration for resources created in TokenServiceStack
--- a/config/constants.ts
+++ b/config/constants.ts
@@ -63,7 +63,7 @@ export const eventBusName = 'OrcaBusMain';
 export const eventSourceQueueName = 'orcabus-event-source-queue';
 
 // Event Archiver constants for EventBus Contruct in SharedStack
-export const archiveBucketNameSuffix = 'orcabus-event-archive-bucket';
+export const archiveBucketNamePrefix = 'orcabus-event-archive-';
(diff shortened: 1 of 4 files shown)


config/stacks/shared.ts
```diff
@@ -44,6 +44,7 @@ const getEventBusConstructProps = (): EventBusProps => {
     vpcProps: vpcProps,
     lambdaSecurityGroupName: archiveSecurityGroupName,
     archiveBucketName: archiveBucketName,
+    enableBucketRetainPolicy: n === 'prod',
```
williamputraintan, line 47: Just a suggestion, perhaps if we could have it as switch-case pattern and pass the RemovalPolicy from the props would be better (e.g the `getEventSourceConstructProps`)? I think the one liner `if-else` is a bit tedious to read for a config file that splits across stages?

williamputraintan (approved):

config/stacks/shared.ts
raylrui: fixed

config/constants.ts
```diff
@@ -63,8 +63,8 @@ export const eventBusName = 'OrcaBusMain';
 export const eventSourceQueueName = 'orcabus-event-source-queue';
 
 // Event Archiver constants for EventBus Contruct in SharedStack
-export const archiveBucketName = 'event-archive-bucket';
```
williamputraintan, line 66: Perhaps we should put the app namespace (`orcabus`) here as well, so #233 does not happen again?

raylrui: added, have a look now

victorskl: @reisingerf must review bucket naming

config/stacks/shared.ts
```diff
@@ -32,19 +32,41 @@ const getSchemaRegistryConstructProps = (): SchemaRegistryProps => {
...
+        bucketRemovalPolicy: RemovalPolicy.DESTROY,
+      };
+    case 'prod':
+      return {
+        ...baseConfig,
+        addCustomEventArchiver: true,
+        archiveBucketName: 'umccr-prod-' + archiveBucketNameSuffix,
+        bucketRemovalPolicy: RemovalPolicy.RETAIN,
+      };
+  }
```
victorskl, line 69: @reisingerf must review bucket naming

victorskl (approved):
Not major of comment from me. 

Better wait from @reisingerf approval on bucket naming before merging. There is some naming convention attach to this. He can advice/guide better.

config/constants.ts
reisingerf: We usually follow the convention to add an account suffix, e.g. `orcabus-event-archive-dev`
I also think we can remove the `-bucket` part (it's the name for a bucket, so that's implicit).

config/stacks/shared.ts
reisingerf: See comment above.

config/constants.ts
raylrui: @reisingerf sure, bucket name convention fixed with prefix + account suffix, `orcabus-event-archive-dev`, `orcabus-event-archive-stg`, `orcabus-event-archive-prod`

victorskl: Hold on. Can we have a quick call for this, pls.? 

Is it account suffix or prefix? And, whether to use AccountID or Account Name alias.

Just an example:
```
12345678901-orcabus-event-archive   --> dev
12345678902-orcabus-event-archive   --> stg
12345678903-orcabus-event-archive   --> prod
```

raylrui: sure, we can have a call. 
do we need AccountID as the prefix?  as the env 'dev', 'stg', and 'prod' suffix already show env.

williamputraintan (approved):

config/constants.ts
raylrui: bucket name has been converted to `<accountId>-orcabus-event-archive` as we discussed.

victorskl: Hang in there, Ray. 

Please be patient with us. We are improving this bucket naming tinkering as we goes.! Your work here is vital & valuable for this move.

We are getting there... :)

raylrui: Great thanks. Happy to contribute and will update the naming when is ready.

config/constants.ts
```diff
@@ -63,8 +77,8 @@ export const eventBusName = 'OrcaBusMain';
...
-export const archiveSecurityGroupName = 'OrcaBusEventArchiveSecurityGroup';
+export const archiveBucketNamePrefix = 'orcabus-event-archive-';
```
reisingerf, line 80: Hm, a quite generic name for a specific purpose bucket. There may be other archive buckets besides the event archive... 

Also, I am not sure this bucket is meant to be for anything else than internal use. Nobody else needs to know that name and it does not have to be configurable. As such we could move its definition into the stack (or even construct) level?

reisingerf (approved):

config/constants.ts
raylrui: I move the Event Archiver definition from into the EventBus Stack level config, as event bus users may need config archiver

PR #255, opened 2024-04-28
Disable Postgres Manager 

williamputraintan:
Error on PostgresManager install will disable this so pipeline could run back.


victorskl:
Early days...

PR #264, opened 2024-05-01
Fix: PG-Manager Build Install

williamputraintan:
Move away from NodeJs cdk lambda to manual nodejs lambda build.

It seems the dependency is not auto added despite pointing to the correct lockfile (I think it uses the top root CDK dependency?). A suggestion [here](https://www.github.com/aws/aws-cdk/issues/14823#issuecomment-1008296914) to handle using command hook but the nodeJs build will change the lockfile, so I would just use the original Lambda construct.


williamputraintan:
Resolve #255

williamputraintan:
🤞

Issue #257, opened 2024-04-30
Review and finalise event schema


victorskl:
## Context

Current we are following EventBridge doc to authoring our event schema.

- https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events-structure.html
- https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-schema-create.html
  - Following example: `StepFunctionsExecutionStatusChange` ^^

We are also empirically observing from `Amazon Console > EventBridge > Schemas > AWS event schema registry` on various AWS built-in schemas.

We are also choosing OpenAPI format over JSONSchema as OpenAPI format is observed to be most use format in built-in registry.

These schema has `AWSEvent` envelope object that wrap the domain event such as `StepFunctionsExecutionStatusChange` or, our case `SequnceRunStateChange` and so on.

After the event schema is put into registry, we can download the corresponding code binding through EventBridge Console UI.
- https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-schema-code-bindings.html

## Forces

There are few concerns we observed so far. As follows.

- Validation: Is EventBridge validate against corresponding schema registry on those messages enter into the bus? Especially archive events, for example. 
- The `detail-type` is a string? Such that:
```
...
"detail-type": "StepFunctionsExecutionStatusChange"
...
```
- What is the use of AWS x-headers? e.g.
```
"AWSEvent": {
  ...
  "x-amazon-events-detail-type": "Step Functions Execution Status Change",
  "x-amazon-events-source": "aws.states",
  ...
```
- Case issue: see our [poll](https://umccr.slack.com/archives/C03ABJTSN7J/p1713416590826269) - favour `camelCase`
- However, note also the envelope level object use `PascalCase`; such as `AWSEvent` and `StepFunctionsExecutionStatusChange` in detail-type property.

## CloudEvent

Study CloudEvent spec (https://github.com/cloudevents/spec) which seem to be well thought-out design.

Its SDK seems to be supported more language toolchains.
https://github.com/cloudevents/spec?tab=readme-ov-file#sdks

Its primer case study article that perform analysis across cloud events status quo.
https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/primer.md

TL;DR example: _JSON Event Format_

https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/json-format.md

```
{
    "specversion" : "1.0",
    "type" : "com.example.someevent",
    "source" : "/mycontext",
[...]

PR #266, opened 2024-05-03
Refactored and deployed event schema stack

victorskl:
* Finalised SequenceRunStateChange event schema - trim & case
* Relocated event schemas directory into docs
* Deployed schema stack

Related #257


victorskl:
I will need to get merge this refactor PR first, before following up next for workflow event schema. At the same time, I need your input and review. So, will be waiting.

williamputraintan (approved):
Hmmm, I just have the preference of having as close as possible to the code so perhaps stay as it is in the config folder 😅 (or maybe at the stack folder) ? I think the `docs/` should be just a bunch of explanation on how to do what instead of relying some actual content. 

Or it might be just me not used to it, so I'm happy either way 😁

config/constants.ts
```diff
@@ -76,7 +76,7 @@ export const rdsMasterSecretName = 'orcabus/master-rds'; // pragma: allowlist se
 validateSecretName(rdsMasterSecretName);
 
 // Other constants that exist in the SharedStack
-export const regName = 'OrcaBusSchemaRegistry';
+export const regName = 'orcabus.events';
```
reisingerf, line 79: I don't think we have other registries planned, but can we be a bit more specific?
E.g. `eventRegName` or `schemaRegName`

docs/event-schemas/srm/SequenceRunStateChange.json
```diff
@@ -56,48 +54,44 @@
...
         "required": [
           "id",
-          "instrument_run_id"
+          "instrumentRunId",
+          "runVolumeName",
+          "runFolderPath",
+          "runDataUri",
+          "sampleSheetName",
+          "startTime",
+          "endTime",
```
reisingerf, line 63: You are keeping one record for each SequenceRun, right?
(i.e. there are defined start/end times and no intermediate state-bound times)

reisingerf (approved):
Minor comments. 
Otherwise LGTM

config/constants.ts
victorskl: Ok, I will change it to `schemaRegistryName`

docs/event-schemas/srm/SequenceRunStateChange.json
victorskl: Yes, that right. One record for each run; unique by `instrumentRunId`.

config/constants.ts
victorskl: Fixed with https://github.com/umccr/orcabus/commit/efd9a5b5fac29a669c8d34e7e4b49dd692585f49

victorskl:
> preference of having as close as possible to the code

Good point, Will. Schema is a bit tricky. Schema is a contract documentation by definition. Putting under docs section emphasizes this point. We might probably using some doc generator on these schemas to present them to user or observer, at some point down track. So, these folks won't need to dig through our code base, too much.

victorskl:
Alexis; Ray; going to merge this now. Comment if any.

Issue #265, opened 2024-05-02
Make upstream service sscheck API base URL lookup environment neutral


victorskl:
**Context:**

Observe the helper code lines around:

https://github.com/umccr/orcabus/blob/d03d6ad/lib/workload/stateless/stacks/bs-runs-upload-manager/layers/src/bs_runs_upload_manager_tools/utils/portal_helpers.py#L28-L29

**Change requests:**

When we deploy to another environments (stg, prod, etc), sscheck API base URL lookup should be discovered accordingly.

**Approaches:**

- We traditionally do this over SSM parameter coordinate lookup to discovery upstream services.
- We can choose to do with config constants mechanic.
- CloudMap is another. We do not have this infra yet. See discussion around [here](https://umccr.slack.com/archives/C03ABJTSN7J/p1714381660323079).

**Actions:**

- Investigate solutions and fix

victorskl:
@reisingerf This one could be critical path to prod deployment.

victorskl:
See also early notes on Trello card. 

https://trello.com/c/QcWZq6U8/1538-orcabus-v1-implement-bclconvert-workflow-manager-microservice-app

PR #204, opened 2024-04-09
icav2 event translator

raylrui:
# Feature/ ica event translator

## Backgrounds
Due to constraints within the AWS Pipes framework, we can't overriding or transform attributes for the ica pipe event easily. This PR introduces a Lambda function named `icav2_event_translator` as a post-processing step. This function will translate events after they are published to the EventBus, modifying/adding necessary fields to conform to our predefined schema orcabus.wfm@WorkflowRunStateChange, and then republish the events with the desired attributes.

### Receive and Translate ICAv2 event to Orcabus internal event
Combine the iput and anylsis result to generat the orcabus internal event.
Event Pattern (filter rules) for event pipe:
```json5
{
        account: [Stack.of(this).account],
        detail: {
          'ica-event': {
            eventCode: ['ICA_EXEC_028'],
            projectId: [{ exists: true }],
            payload: {
              pipeline: {
                code: [{ prefix: 'BclConvert' }],
              },
            },
          },
        },
},
```

### Save this translation process to dynamoDB
Send the orignal input and orcabus event after translation to the dynao db for recording of the process. 

### Publish the Orcabus event to event bus
publish the Orcabus (internal) event back the event bus.
Example output event:
```json5
{
  "version": "0", // by default is 0
  "id": "dxxxxx-6xxxx-fxxx-1xxx-5xxxxx",
  "detail-type": "WorkflowRunStateChange", // event type
  "source": "orcabus.bcm", // from bct service
  "account": "404687356768",
  "time": "2024-00-00T02:08:46Z",
  "region": "ap-southeast-x",
  "resources": [],
  "detail": {
    // translated event details
  }
}
```

PR #270, opened 2024-05-06
Fix ica event pipe name

raylrui:
Fix the ica event pipe name with the fixed pipe name.

So in the event rules pattern, we can use `"source": "Pipe IcaEventPipeStackPipe",` rather than aws assigned id ` "source": "Pipe IcaEventPipeConstru-aVEAl34nl7zf",`

### DIFF
--- a/lib/workload/stateful/stacks/ica-event-pipe/constructs/ica_event_pipe/index.ts
+++ b/lib/workload/stateful/stacks/ica-event-pipe/constructs/ica_event_pipe/index.ts
@@ -36,7 +36,7 @@ export class IcaEventPipeConstruct extends Construct {
     super(scope, id);
     this.icaQueue = this.createMonitoredQueue(id, props).queue;
     this.mainBus = EventBus.fromEventBusName(this, 'EventBus', props.eventBusName);
-    this.createPipe();
+    this.createPipe(props.icaEventPipeName);
   }
 
   // Create the INPUT SQS queue that will receive the ICA events
@@ -46,7 +46,7 @@ export class IcaEventPipeConstruct extends Construct {
     // However, our use case, as we don't add any additional subscriptions, does not require topic modification, so we can pass on an "ITopic" as "Topic".
     const topic: Topic = Topic.fromTopicArn(this, 'SlackTopic', props.slackTopicArn) as Topic;
 
-    const mq = new MonitoredQueue(this, props.icaEventPipeName, {
+    const mq = new MonitoredQueue(this, props.icaQueueName, {
       queueProps: {
         queueName: props.icaQueueName,
         enforceSSL: true,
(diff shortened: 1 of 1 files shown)


reisingerf:
@williamputraintan @victorskl 
FYI: I guess we'll be running into those kind of issues... and so far there are a couple of options I could see as a solution, please add if you have others:

We don't want direct stack dependencies, so need a different way to allow entity resolution. 

Fixing entity names in the shared config is one option, but may come with limitations on deployability (e.g. creating unique independent entities). Slightly better may be to use name prefixes (with autogenerated suffix) to ensure unique names, but that may create issues with the discovery process (lookup using wildcards or filtering if more than one resource).

I guess a better option would be to register resources that are shared into a discovery service (like CloudMap) so they are available to consumers within the correct context.

I think for now I am OK with fixing that name.

victorskl:
Okie. So, this is pinning the named resource so that there shouldn't be two instances of the same should exist, correct?

victorskl:
> discovery service (like CloudMap) 

See another case #265

reisingerf:
Yes, it has popped up before... 
CloudMap could be a good option, but until we have explored and instrumented that a bit more I am happy with fixing the names

raylrui:
Thanks for your comment.
My first thought about this is just put the `props.icaEventPipeName` in the new pipe resource, so the event pipe we create here has name define as in the props, like this
```
new pipes.Pipe(this, "Pipe", {
      pipeName: props.icaEventPipeName,
      source: new SqsSource(this.icaQueue),
      target: new EventBusTarget(this.mainBus, {
        inputTransformation: targetInputTransformation,
      }),
```
And then, in my translator service, I can use the eventPattern with define pipe source like this 😅:
```
const rule = new Rule(this, 'Rule', {
      eventBus: this.mainBus,
      eventPattern: {
        account: [Stack.of(this).account],
        detailType: ['Event from aws:sqs'],
        source: [`Pipe ${props.IcaEventPipeName}Pipe`],
        detail: {
          'ica-event': ['*'],
        },
      },
    });
```
I am not sure if it bring too much here. But sure, CloudMap will be the great option to map all service rather than define fix names in our constants.

victorskl:
That's ok. Good discussion. Early days. Let us keep moving. This is fine. We can flex this, if need be.

victorskl:
Btw, event `source` naming follows reverse domain naming pattern. Like Java package naming.

```
orcabus.<service1>
orcabus.<service2>
orcabus.<service3>
```

Not sure; it is viable to maintain this pattern here. If we can, perhaps `orcabus.icapipe` might be good candidate...

raylrui:
>Not sure; it is viable to maintain this pattern here. If we can, perhaps orcabus.icapipe might be good candidate...

As I research, AWS Pipes handles transformations, especially for critical fields like `source` and `detail-type` in EventBridge events. These fields are core attributes of the event itself, managed at a higher level than the `detail` payload, which is what is typically manipulated by transformations in AWS Pipes. 
So AWS Pipes **doesn't directly allow overriding the source and detail-type at the Pipe level** because these are set by the AWS Pipes framework itself when routing events through Pipes.

raylrui:
Hi @reisingerf @victorskl  I will drop the PR for now, as I have coupled the translator service within the ica event pipe stack in another PR, see #204.
For the name convention, we may have more PRs in the future.

Issue #103, opened 2024-02-12
Implement workflow manager - baseline


victorskl:
**Context:**

* Initialise workflow manager app following Django/Python
* Setup local dev and compose, etc

victorskl:
Before kicking off with the Django (as current [proposed tech stack](https://github.com/umccr/orcabus/blob/5b2b5b4/docs/developer/DJANGO_API.md) of choice), allow me to explore the tech and study about:

- ASGI
- FastAPI
- mono-Lambda (proxy+) Vs single-responsible Lambda (API routing pattern)

Also ditto Slack [standup](https://umccr.slack.com/archives/C025SCF86/p1708987995109089) daily.

victorskl:
Evaluation is in.
- will use ASGI integration (over WSGI)
- will use mono-Lambda reverse proxy pass pattern
- will continue stick around with Django/DRF for now

victorskl:
Service namespace: `orcabus.wfm`

PR #272, opened 2024-05-06
Implemented WorkflowRunStateChange schema

victorskl:
* Use opaque Payload object for differing workflow type
* Introduce two workflow types: `bssh_icav2_fastq_copy` and `bssh_bcl_convert`
* Introduce 3 namespaces for services: `orcabus.wfm`, `orcabus.bct`, `orcabus.fqcp`

### DIFF
--- a/docs/event-schemas/wfm/WorkflowRunStateChange.json
+++ b/docs/event-schemas/wfm/WorkflowRunStateChange.json
@@ -12,9 +12,7 @@
         "required": [
           "detail-type",
           "detail",
-          "id",
-          "source",
-          "time"
+          "source"
         ],
         "properties": {
           "detail": {
@@ -80,21 +78,22 @@
           },
           "payload": {
             "$ref": "#/components/schemas/Payload"
-          },
-          "serviceVersion": {
-            "type": "string"
           }
         }
       },
       "Payload": {
         "type": "object",
         "required": [
-          "refId"
+          "refId",
+          "version"
         ],
         "properties": {
           "refId": {
             "type": "string",
             "nullable": true
+          },
+          "version": {
+            "type": "string"
           }
         },
         "additionalProperties": true
--- a/docs/event-schemas/wfm/WorkflowRunStateChange.json
+++ b/docs/event-schemas/wfm/WorkflowRunStateChange.json
@@ -1,0 +1,104 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "version": "1.0.0",
+    "title": "WorkflowRunStateChange"
+  },
+  "paths": {},
+  "components": {
+    "schemas": {
+      "AWSEvent": {
+        "type": "object",
+        "required": [
+          "detail-type",
+          "detail",
+          "id",
+          "source",
+          "time"
+        ],
+        "properties": {
+          "detail": {
+            "$ref": "#/components/schemas/WorkflowRunStateChange"
+          },
+          "account": {
+            "type": "string"
+          },
+          "detail-type": {
+            "type": "string"
+          },
+          "id": {
+            "type": "string"
+          },
+          "region": {
+            "type": "string"
+          },
+          "resources": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          },
+          "source": {
+            "type": "string"
+          },
+          "time": {
+            "type": "string",
+            "format": "date-time"
+          },
+          "version": {
+            "type": "string"
+          }
+        }
+      },
+      "WorkflowRunStateChange": {
+        "type": "object",
+        "required": [
+          "portalRunId",
+          "timestamp",
+          "status",
+          "workflowType",
+          "workflowVersion",
+          "payload",
+          "serviceVersion"
+        ],
+        "properties": {
+          "portalRunId": {
+            "type": "string"
+          },
+          "timestamp": {
+            "type": "string",
+            "format": "date-time"
+          },
+          "status": {
+            "type": "string"
+          },
+          "workflowType": {
+            "type": "string"
+          },
+          "workflowVersion": {
+            "type": "string"
+          },
+          "payload": {
+            "$ref": "#/components/schemas/Payload"
+          },
+          "serviceVersion": {
+            "type": "string"
+          }
+        }
+      },
+      "Payload": {
+        "type": "object",
+        "required": [
+          "refId"
+        ],
+        "properties": {
+          "refId": {
+            "type": "string",
+            "nullable": true
+          }
+        },
+        "additionalProperties": true
+      }
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


victorskl:
I still feel uneasy about Payload object being opaque about its attributes. It is practically making it un-controlled schema or domain object. 

Mechanism such as Schema Registry and Code binding (auto-generated code) missed this opaqueness. Hence, this would effectively dodge the schema control radar. If the upstream producer started to divergent from the initial payload structure, the subscriber may not aware of this. And, it'd ends up manual handle on event parsing.

But. Having said, let experiment this. This is what we discussed the other day.

victorskl:
I have added [reading pointers](https://github.com/umccr/orcabus/issues/257#issuecomment-2095605777) to ticket. If you are able, highly recommended to give a read.

victorskl:
Next, I will start prototyping baseline wfm #103 -- based on this WRSC. FYI

raylrui (approved):
LGTM!

victorskl:
Self note on design: 

This is fine as we treat the all of Workflow business as one domain aggregate (DDD bounded context). Then, these are all internal events to this bounded context of the Workflow world and, events happening within it.

We can come up with another dedicated schema down the track; when externally observer would like to communicate into Workflow orchestration world events.

docs/event-schemas/wfm/WorkflowRunStateChange.json
```diff
@@ -0,0 +1,104 @@
...
+  "components": {
+    "schemas": {
+      "AWSEvent": {
+        "type": "object",
+        "required": [
+          "detail-type",
+          "detail",
+          "id",
+          "source",
+          "time"
```
reisingerf, line 17: Should we make `account` mandatory? It's used in the archive filter.
Also, latest once we fix the payload schema(s), we should make the version mandatory.

docs/event-schemas/wfm/WorkflowRunStateChange.json
```diff
@@ -0,0 +1,104 @@
...
+      "WorkflowRunStateChange": {
+        "type": "object",
+        "required": [
+          "portalRunId",
+          "timestamp",
+          "status",
+          "workflowType",
+          "workflowVersion",
+          "payload",
+          "serviceVersion"
```
reisingerf, line 62: Do we need a `service` or is the `source` sufficient?

reisingerf (approved):

docs/event-schemas/wfm/WorkflowRunStateChange.json
alexiswl: What's the difference between service and source?

alexiswl:
> I still feel uneasy about Payload object being opaque about its attributes. It is practically making it un-controlled schema or domain object.

I agree but I don't think this is 'forever', good to get started with something and we can move to something more robust later. 

I think once we're up and running with our set services we can move this to something like

```
{
  "payload": {
    "type": "object",
   "properties": "id",
   "oneOf": [
      {
        "$ref": "#definitions/microserviceA"
      },
      {
        "$ref": "#definitions/microserviceB"
      }
   ]
  }
}
```

victorskl:
Yup. Thought about that too, Alexis.

docs/event-schemas/wfm/WorkflowRunStateChange.json
victorskl: I think Flo is referring to `serviceVersion`

docs/event-schemas/wfm/WorkflowRunStateChange.json
victorskl: Yup. We can do that.

docs/event-schemas/wfm/WorkflowRunStateChange.json
victorskl: > The service differentiator marker on events is the `source` property of the event envelope.

docs/event-schemas/wfm/WorkflowRunStateChange.json
victorskl: Final AWS Envelope:

We did run experiment and, turned out `id` and `version` are not in our control. EventBus fills these value regardless. Also, `version` seems to be always set to `0` on the actual event instance that sent into the channel; though we payload it with `1.0.0`. 

The `account`, `region`, `time` observed to be auto-filled by EventBridge as well.

Hence. We concluded that we will be just doing the required fields on `detail`, `detail-type`, `source` as per AWS documentation. See below.

Minimum information needed for a valid custom event
https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-events-structure.html

docs/event-schemas/wfm/WorkflowRunStateChange.json
victorskl: We discussed about pros/cons of opaqueness of Payload object. And, tried tinker with alternate approaches such as fixing the schemas upfront.

For now, we concluded we shall keep it as Payload object.

The `serviceVersion` will be refactored to `version` property of the `Payload` definition as the required field; along with payload `refId`.

docs/event-schemas/wfm/WorkflowRunStateChange.json
raylrui: And some more reading on this:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html
The version seems applied in "Create a schema from events on an event bus"
>If the contents of events on the event bus change, EventBridge creates new versions of the related EventBridge schema.
https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-schema-create.html

victorskl: Updated with https://github.com/umccr/orcabus/commit/1e903c67c1eb0fe8ffa949171a058c93c835c4d7

docs/event-schemas/wfm/WorkflowRunStateChange.json
victorskl: Updated with https://github.com/umccr/orcabus/commit/1e903c67c1eb0fe8ffa949171a058c93c835c4d7

victorskl:
@alexiswl @williamputraintan tick pls if you are ok with PR

alexiswl (approved):

williamputraintan (approved):

docs/event-schemas/wfm/WorkflowRunStateChange.json
victorskl: Also harmonised/updated SRSC envelope with https://github.com/umccr/orcabus/commit/34904419b98e20498a02a65c685c93f2ba93993d

victorskl:
Alright. Good luck with our very first schema.! 🤞 This is the start...

PR #281, opened 2024-05-08
Use 'currentVersion' for lambda object to resolve the cdk nag issue

alexiswl:
Related issue: https://github.com/aws/aws-cdk/issues/20177

`grantInvoke` will use `*` for all lambda versions which breaks cdk-nag.  

By using `currentVersion` attribute of the lambda obj we only grantInvoke for only the latest version of the lambda object, resolving the cdc-nag errors

### DIFF
--- a/lib/workload/stateless/stacks/cttso-v2-pipeline-manager/deploy/constructs/cttsov2-icav2-manager/index.ts
+++ b/lib/workload/stateless/stacks/cttso-v2-pipeline-manager/deploy/constructs/cttsov2-icav2-manager/index.ts
@@ -194,7 +194,7 @@ export class Cttsov2Icav2PipelineManagerConstruct extends Construct {
       generate_trimmed_samplesheet_lambda_obj,
       upload_samplesheet_to_cache_dir_lambda_obj,
     ].forEach((lambda_obj) => {
-      lambda_obj.grantInvoke(<iam.IRole>stateMachine.role);
+      lambda_obj.currentVersion.grantInvoke(<iam.IRole>stateMachine.role);
     });
 
     // Allow state machine to read/write to dynamodb table


victorskl (approved):
👍

lib/workload/stateless/stacks/cttso-v2-pipeline-manager/deploy/constructs/cttsov2-icav2-manager/index.ts
```diff
@@ -194,7 +194,7 @@ export class Cttsov2Icav2PipelineManagerConstruct extends Construct {
       generate_trimmed_samplesheet_lambda_obj,
       upload_samplesheet_to_cache_dir_lambda_obj,
     ].forEach((lambda_obj) => {
-      lambda_obj.grantInvoke(<iam.IRole>stateMachine.role);
+      lambda_obj.currentVersion.grantInvoke(<iam.IRole>stateMachine.role);
```
victorskl, line 197: @williamputraintan @mmalenic @raylrui FYI, folks. 

If we use `currentVersion` prop of Lambda, we can get away of cdk generating `*` to ARN. This is for cdk-nag `AwsSolutions-IAM5` suppression.

reisingerf (approved):

alexiswl:
Ah this doesn't work as magically as one would have hoped. 

Using current version is very specific - 
Got this error in AWS Step functions

```
User: arn:aws:sts::843407916570:assumed-role/OrcaBusBeta-BclconvertInt-BclconvertInteropQcIcav2P-olRiObau8Dle/OfpsBsQSvUGVcbvTGeepNuvQApaXaOts is not authorized to perform: lambda:InvokeFunction on resource: arn:aws:lambda:ap-southeast-2:843407916570:function:OrcaBusBeta-BclconvertInt-BclconvertInteropQcIcav2-WOdb9Nfb3N0C because no identity-based policy allows the lambda:InvokeFunction action (Service: AWSLambda; Status Code: 403; Error Code: AccessDeniedException; Request ID: 04b41406-99a6-489a-b09a-2a0224262b21; Proxy: null)
```

Because step functions role has access to `arn:aws:lambda:ap-southeast-2:843407916570:function:OrcaBusBeta-BclconvertInt-BclconvertInteropQcIcav2-WOdb9Nfb3N0C:1` not `arn:aws:lambda:ap-southeast-2:843407916570:function:OrcaBusBeta-BclconvertInt-BclconvertInteropQcIcav2-WOdb9Nfb3N0C`.  

So when replacing definitions in step functions, rather than `lambda_obj.functionArn`, need to do `lambda_obj.currentVersion.functionArn`

victorskl:
Argh. CDK meow.! 😿 Must be reason, why CDK grant generate slash star it.

Go for it, Alexis. Suppress `AwsSolutions-IAM5` nag like other would. All good.

https://github.com/umccr/orcabus/blob/3638b2c/test/stateless/deployment.test.ts#L131-L189

reisingerf:
Is it an issue passing the `lambda_obj.currentVersion.functionArn` to the step function?

alexiswl:
No it's not, it's just something to remember.  

It does though, mean that everytime we update the lambda and deploy (manually), cdk will ask for permission first since we'll also be updating the role permissions of the step function

PR #282, opened 2024-05-08
ICAv1 copy batch utility

brainstorm:
This PR should, eventually:

1. Copy all files from gds://production
2. Have Filemanager consume the events that the copied objects generate, without having to manually inject entries on the database.
3. Compute hashes that can be compared validation on target destination.
4. Store this information in the filemanager so that object e_tags can be compared, e.g. each object could store e_tag and chunk size pairs. In addition, keeping in mind that the e_tag is relevant with a corresponding chunk size (for objects larger than a particular size). So there would probably be multiple Etags stored (ETag-8MiB, ETag-16MiB), depending on uploader client (aws cli, boto3, etc..) defaults and overrides.


config/constants.ts
```diff
@@ -99,6 +99,24 @@ Resources required for BaseSpace TES stack
 
 export const basespaceAccessTokenSecretName = '/manual/BaseSpaceAccessTokenSecret'; // pragma: allowlist secret
 
+/*
+ICAv1 Resources
+ */
+export const icav1AccessTokenSecretName: Record<AppStage, string> = {
+  [AppStage.BETA]: 'ICAv1JWTKey-umccr-prod-service-trial', // pragma: allowlist secret
```
alexiswl, line 106: I don't think these keys exist,

lib/workload/stateless/stacks/icav1-copy-batch-utility/lambdas/lambda.py
```diff
@@ -0,0 +1,184 @@
...
+            tagging_to_s3 = "&".join([f"{parse.quote_plus(d['Key'])}={parse.quote_plus(d['Value'])}" for d in existing_tag_set])
+            # Construct Request With Required and Available Arguments
+            if existing_tag_set:
+                myargs['Tagging'] = tagging_to_s3
+        else:
+            logger.info("Object Tagging Copy Disabled")
+
+        # Initiate the Actual Copy Operation and include transfer config option
+        logger.info(f"starting copy of object {s3Key} with versionID {s3VersionId} between SOURCEBUCKET: {s3Bucket} and DESTINATIONBUCKET: {newBucket}")
+        response = s3Client.copy(copy_source, newBucket, newKey, Config=transfer_config, ExtraArgs=myargs)
```
alexiswl, line 134: How long would this be expected to take for a 100 Gb file?

alexiswl:
Could also do with a Readme, what are the inputs into the lambda?

Also wondering, what's the rationale for pulling this into the orcabus if this is a 'one-off'?

Could this be its own standalone cdk?

config/constants.ts
brainstorm: No, they don't, draft still, working on it.

brainstorm:
> Could also do with a Readme, what are the inputs into the lambda?
> 

Yep, WIP, WIP, WIP. Inputs would be mainly an ICAv1 gds volume in its last iteration I reckon.

> 
> Also wondering, what's the rationale for pulling this into the orcabus if this is a 'one-off'?
> 


I thought about having it as an independent repo, but this way it gets more eyes/scrutiny (after it comes out of draft status), which is needed IMHO… might need to interact with FileManager perhaps? Other unforeseen integration gotchas.


> 
> Could this be its own standalone cdk?
> 
> 


Yes, ditto above.

lib/workload/stateless/stacks/icav1-copy-batch-utility/lambdas/lambda.py
brainstorm: We'll see after implementing this: https://aws.amazon.com/blogs/storage/copying-objects-greater-than-5-gb-with-amazon-s3-batch-operations/

Issue #285, opened 2024-05-08
Orcabus stateless codepipeline fails to deploy with: `Artifact [Synth_Output] exceeds max artifact size`


alexiswl:
Some suggestions by William The CodeWhisperer: 
https://github.com/aws/aws-cdk/issues/9917
https://github.com/aws/aws-cdk/issues/21325

Specifically https://github.com/aws/aws-cdk/issues/21325#issuecomment-1667180293 may help us

PR #286, opened 2024-05-08
Remove cdk.out assets to prevent synth output

alexiswl:
Resolves https://github.com/umccr/orcabus/issues/285 (in theory)


williamputraintan (approved):
Happy to try this first, else the removal needs to be from a different stage.

PR #287, opened 2024-05-08
Revert "Remove cdk.out assets to prevent synth output"

alexiswl:
This reverts commit 2d20a9df850d44cfac90274aa2f1091ad2edb39f.


alexiswl:
@williamputraintan did you want to give this a test run?

alexiswl:
Ah wait, I see that's happened, FileManagerStack now fails

williamputraintan:
Yes, did run this manually. But now fails on different part, will investigate on it.

williamputraintan:
Will try let the pipeline handle the build instead of manual deployment.

Issue #100, opened 2024-02-08
initiate github action for unit testing


williamputraintan:
to enable fast feedback if unit test fail before merging to main branch

victorskl:
Also to investigate GitHub <> CodeBuild integration. Fairly new, but it is valuable if this work well.

- https://aws.amazon.com/about-aws/whats-new/2023/07/aws-codebuild-github-actions/
- https://docs.aws.amazon.com/codebuild/latest/userguide/action-runner.html
- https://www.google.com/search?q=codebuild+github+action

PR #290, opened 2024-05-10
feat: Refine Testing Configuration

williamputraintan:
Put back GH Action which runs on AWS CodeBuild as GH Runners except for lint and Security check. All application test runs at different CodeBuild runner to cut on time.

### DIFF
--- a/lib/pipeline/statelessPipelineStack.ts
+++ b/lib/pipeline/statelessPipelineStack.ts
@@ -37,7 +37,7 @@ export class StatelessPipelineStack extends cdk.Stack {
       },
       environment: {
         type: 'ARM_CONTAINER',
-        computeType: 'BUILD_GENERAL1_SMALL',
+        computeType: 'BUILD_GENERAL1_LARGE',
         image: 'aws/codebuild/amazonlinux2-aarch64-standard:3.0',
         privilegedMode: true,
       },
--- a/lib/pipeline/statelessPipelineStack.ts
+++ b/lib/pipeline/statelessPipelineStack.ts
@@ -28,7 +28,7 @@ export class StatelessPipelineStack extends cdk.Stack {
     });
     new codebuild.CfnProject(this, 'GHRunnerCodeBuildProject', {
       // the name here act as a unique id for GH action to know which CodeBuild to use
-      // So if you change this, you need to update the GH action yaml file (.github/workflows/prbuild.yml)
+      // So if you change this, you need to update the GH action .yml file (.github/workflows/prbuild.yml)
       name: 'orcabus-codebuild-gh-runner',
       description: 'GitHub Action Runner in CodeBuild for `orcabus` repository',
       serviceRole: ghRunnerRole.roleArn,
--- a/lib/pipeline/statelessPipelineStack.ts
+++ b/lib/pipeline/statelessPipelineStack.ts
@@ -20,6 +20,41 @@ export class StatelessPipelineStack extends cdk.Stack {
   constructor(scope: Construct, id: string, props: cdk.StackProps) {
     super(scope, id, props);
 
+    // Define CodeBuild project for GH action runner to use
+    // the GH repo defined below already configured to allow CB webhook
+    // This is actually not part of the pipeline, so I guess we could move this someday.
+    const ghRunnerRole = new iam.Role(this, 'GHRunnerRole', {
+      assumedBy: new iam.ServicePrincipal('codebuild.amazonaws.com'),
+    });
+    new codebuild.CfnProject(this, 'GHRunnerCodeBuildProject', {
+      // the name here act as a unique id for GH action to know which CodeBuild to use
+      // So if you change this, you need to update the GH action yaml file (.github/workflows/prbuild.yml)
+      name: 'orcabus-codebuild-gh-runner',
+      description: 'GitHub Action Runner in CodeBuild for `orcabus` repository',
+      serviceRole: ghRunnerRole.roleArn,
+      artifacts: {
+        type: 'NO_ARTIFACTS',
+      },
+      environment: {
+        type: 'ARM_CONTAINER',
+        computeType: 'BUILD_GENERAL1_SMALL',
+        image: 'aws/codebuild/amazonlinux2-aarch64-standard:3.0',
+        privilegedMode: true,
+      },
+      source: {
+        type: 'GITHUB',
+        gitCloneDepth: 1,
+        location: 'https://github.com/umccr/orcabus.git',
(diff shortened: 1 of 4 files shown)


williamputraintan:
Well, I guess we really need to setup self hosted runners in CodeBuild. GH Runners do not support arm architecture 😐

victorskl:
Go for it, Will. CodeBuild runner sounds good. 

At least we will learn a thing or two for trying it. ACK; it is a new feature between GitHub Action and AWS CodeBuild side of eco-system. 

Plus, if it works well, no worries about potential split billing issue about compute cost on building the software stacks. Compute cost will be consolidated at AWS end.

We achieve best of both worlds then.


FYI @ohofmann @reisingerf

victorskl:
Related to #100 story from the Platform engineering backlog

lib/pipeline/statelessPipelineStack.ts
```diff
@@ -20,6 +20,41 @@ export class StatelessPipelineStack extends cdk.Stack {
...
+        image: 'aws/codebuild/amazonlinux2-aarch64-standard:3.0',
+        privilegedMode: true,
+      },
+      source: {
+        type: 'GITHUB',
+        gitCloneDepth: 1,
+        location: 'https://github.com/umccr/orcabus.git',
+        reportBuildStatus: false,
+      },
+      logsConfig: { cloudWatchLogs: { status: 'DISABLED' } },
```
williamputraintan, line 50: Not seeing anything is useful here, might as well disabled this.

Screenshot 2024-05-13 at 17 42 27

lib/pipeline/statelessPipelineStack.ts
```diff
@@ -20,6 +20,41 @@ export class StatelessPipelineStack extends cdk.Stack {
   constructor(scope: Construct, id: string, props: cdk.StackProps) {
     super(scope, id, props);
 
+    // Define CodeBuild project for GH action runner to use
+    // the GH repo defined below already configured to allow CB webhook
+    // This is actually not part of the pipeline, so I guess we could move this someday.
+    const ghRunnerRole = new iam.Role(this, 'GHRunnerRole', {
+      assumedBy: new iam.ServicePrincipal('codebuild.amazonaws.com'),
+    });
+    new codebuild.CfnProject(this, 'GHRunnerCodeBuildProject', {
```
williamputraintan, line 29: Well, this webhook GH runner <> AWS CodeBuild is [new](https://aws.amazon.com/about-aws/whats-new/2024/04/aws-codebuild-managed-github-action-runners/), so is not a surprised that the webhook flag for L2 CDK construct hasn't exist yet. Defining here the L1 construct.

[cdk-issue-#30062](https://www.github.com/aws/aws-cdk/issues/30062)

lib/pipeline/statelessPipelineStack.ts
```diff
@@ -20,6 +20,41 @@ export class StatelessPipelineStack extends cdk.Stack {
...
+      triggers: {
+        webhook: true,
+        buildType: 'BUILD',
+        filterGroups: [[{ type: 'EVENT', pattern: 'WORKFLOW_JOB_QUEUED' }]],
+      },
+    });
+
     // A connection where the pipeline get its source code
     const codeStarArn = ssm.StringParameter.valueForStringParameter(this, 'codestar_github_arn');
     const sourceFile = pipelines.CodePipelineSource.connection('umccr/orcabus', 'main', {
```
williamputraintan, line 60: On the unittest below, did want to try the [batch/build-list](https://docs.aws.amazon.com/codebuild/latest/userguide/batch-build-buildspec.html#build-spec.batch.build-list), so could speed up by parallel testting, but might not be convenient yet for cdk-pipeline codebuild source. a feature request was put to cdk issue [#29592](https://www.github.com/aws/aws-cdk/issues/29592), smth that we could look for in the future.

lib/pipeline/statelessPipelineStack.ts
```diff
@@ -20,6 +20,41 @@ export class StatelessPipelineStack extends cdk.Stack {
...
 
+    // Define CodeBuild project for GH action runner to use
+    // the GH repo defined below already configured to allow CB webhook
+    // This is actually not part of the pipeline, so I guess we could move this someday.
+    const ghRunnerRole = new iam.Role(this, 'GHRunnerRole', {
+      assumedBy: new iam.ServicePrincipal('codebuild.amazonaws.com'),
+    });
+    new codebuild.CfnProject(this, 'GHRunnerCodeBuildProject', {
+      // the name here act as a unique id for GH action to know which CodeBuild to use
+      // So if you change this, you need to update the GH action yaml file (.github/workflows/prbuild.yml)
```
reisingerf, line 31: Huh? 
The name here does not match the name in the prbuild.yml file though. How are they matched up?

lib/pipeline/statelessPipelineStack.ts
reisingerf: Hm, the useful bits may only come when something goes wrong? 
Are those logs visible via the codebuild history? 
In any case, I guess easy to add back in if we need to...

reisingerf (approved):
Let's try!

lib/pipeline/statelessPipelineStack.ts
williamputraintan: The real testing logs goes to the GH action itself.

https://github.com/umccr/orcabus/actions/runs/9063282042

lib/pipeline/statelessPipelineStack.ts
williamputraintan: Ahh so the pattern on the GH action side is below, so the project name should be the exact name. I'll update this on the commenting

```
runs-on: codebuild-<project-name>-${{ github.run_id }}-${{ github.run_attempt }}
```

lib/pipeline/statelessPipelineStack.ts
williamputraintan: To enable just need to change the `DISABLED` to `ENABLED`, and need to add log group (and its respective role) as this is the L1 construct

victorskl (approved):
Good job, Will.

williamputraintan:
cc @alexiswl @brainstorm @mmalenic @raylrui 

Just a side note if you are not aware of the `[skip ci]` flag (when added part of your commit message), it lets you skip the GH Action Jobs. Useful if you know your commit do not break any tests (e.g. code commenting commit). Eventually before it gets deployed, the code will get tested again.

https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

Issue #293, opened 2024-05-15
Deprecate icav2-copy-batch-utility


alexiswl:
ICAv2 Copy Batch Utility as a stack is overkill.  

As @reisingerf suggested, each service that requires the copy batch utility can generate their own.  

Create a copy batch utility as a construct and then have microservices create their own versions of this construct.

victorskl:
_(made me shock momentarily first)_

> Create a copy batch utility as a construct 

Composing it as a construct sound good.! 👍

alexiswl:
It does mean that the cttsov2 service will have five step functions associated with its stack

victorskl:
Hear you. Hang in there. Pls yell, if step functions started to become out of hand. We can always regroup and discuss on tech choices...

brainstorm:
This decision makes sense, in whatever meeting that happened, I've been thinking the same while working on the ICAv1 copy batch utility stack, more generic constructs are needed and I reckon that AWS S3 Batch Ops can do a good job at that.

alexiswl:
>  in whatever meeting that happened

No meeting, just orcabus daydreams in my head.  

One other benefits is that the step functions executions are less cluttered with different services. 
If I want to see when the cttso executed the copy batch I can easily find executions of this without needing to trawl through the bssh fastq copy executions (because they're different step functions now with different names).

PR #192, opened 2024-04-03
Feature/bclconvert interop qc manager (and refactor icav2 configuration)

alexiswl:
# BCLConvert Interop QC CWL pipeline

## TODO 

- [x] Wait for https://github.com/umccr-illumina/ica_v2/issues/180 to test if this pipeline actually works
- [x] Add eventFilter and subscribe to orcabus internal events (such as workflow manager)
- [x] Add eventFilter and subscribe to external events (such as ICAv2 pipeline state change)
- [x] Fix putEvents at the end of each of the step functions to fire at the actual orcabus event pipe - state machine's permissions might need to be adjusted for this to actually work
- [x] Integrate into orcabus codepipeline for deployment
- [ ] Reorg payloads such that all payload items are under the 'data' section, except for 'refId' and 'version'
- [ ] Reorg workflow run statechange inputs so that workflowType is now workflowName, source is longhand version of source name
- [ ] Work with #305 to determine harmonious event communication between systems

## Readmes 

To view the Readmes related to this PR

* [BCLConvert InterOp QC Stack](https://github.com/umccr/orcabus/tree/feature/bclconvert-interop-qc-manager/lib/workload/stateless/stacks/bclconvert-interop-qc-pipeline-manager)
* [ctTSOv2 Stack](https://github.com/umccr/orcabus/tree/feature/bclconvert-interop-qc-manager/lib/workload/stateless/stacks/cttso-v2-pipeline-manager)

* [ICAv2 Handle WFM Ready Event Construct](https://github.com/umccr/orcabus/tree/feature/bclconvert-interop-qc-manager/lib/workload/components/dynamodb-icav2-ready-event-handler-sfn)
* [ICAv2 Handle Event State Change SFN Construct](https://github.com/umccr/orcabus/tree/feature/bclconvert-interop-qc-manager/lib/workload/components/dynamodb-icav2-handle-event-change-sfn)

PR #295, opened 2024-05-15
Add Events system to bssh fastq copy manager (and deprecate icav2 copy batch utility stack)

alexiswl:
Refactor bssh and cttso v2 so each service generates its own icav2 copy file service, rather than use a standalone copy batch utility.  

ICAv2 file copy construct has both single and batch constructs so cttsov2 can use single (since it's only using one dest uri).  

See https://github.com/umccr/orcabus/issues/293 for more info

## TODO 

> ~~This branch is forked from the PR below so this needs to be merged first~~
- [ ] ~~PR #192 Instead we merge into #192~~


alexiswl:
@williamputraintan if I delete a service from the stack will it actually delete? I guess we will find out soon

alexiswl:
Also I will deal with secrets detected by GitGuardian

williamputraintan:
> @williamputraintan if I delete a service from the stack will it actually delete? I guess we will find out soon

I don't think so, might need to `yarn cdk-stateless destroy ...` manually

alexiswl:
> I don't think so, might need to yarn cdk-stateless destroy ... manually

Cool, will wait until this is merged before doing that then, incase there is actually a service in prod using it

PR #296, opened 2024-05-16
Feat: Limit PR Testing

williamputraintan:
As discussed, do not run GH Action test on draft-pr

### DIFF
--- a/.github/workflows/prbuild.yml
+++ b/.github/workflows/prbuild.yml
@@ -3,7 +3,13 @@ name: Pull Request Build
 on:
   pull_request:
     paths-ignore:
+      # The ignore only works only if changes to the main branch only include the following files.
+      # So if the commit only contain .md changes but the PR change contain more, the ignore fails
+      # https://github.com/actions/runner/issues/2324#issuecomment-1703345084
       - '**.md'
+      - '**.svg'
+      - '**.drawio'
+      - '**.png'
     types:
       - opened
       - reopened
--- a/.github/workflows/prbuild.yml
+++ b/.github/workflows/prbuild.yml
@@ -2,6 +2,8 @@ name: Pull Request Build
 
 on:
   pull_request:
+    types:
+      - ready_for_review
     branches:
       - main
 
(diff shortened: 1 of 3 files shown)


victorskl (approved):
LGTM. Thanks, Will.

.github/workflows/prbuild.yml
```diff
@@ -2,6 +2,11 @@ name: Pull Request Build
 
 on:
   pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
     branches:
       - main
```
raylrui, line 11: can we add path filter here in case unnecessary build for docs file?
```
paths:
      - "**"
      - "!/*.md"
      - "!/**.md"
 ```

Comment (victorskl) [reply .github/workflows/prbuild.yml:11]: Good point, Ray! Perhaps, .png, .svg, .drawio too. Just in case.

Comment (williamputraintan) [reply .github/workflows/prbuild.yml:11]: Good idea! Thanks

Comment (williamputraintan) [reply .github/workflows/prbuild.yml:13]: @raylrui Hmm, tried that with your config but doesn't seems to work as expected. Tried updating some `.md` but still run the test. Do you have any idea what went wrong?

Comment (raylrui) [reply .github/workflows/prbuild.yml:13]: sure, I will have a check again

Comment (raylrui) [reply .github/workflows/prbuild.yml:13]: @williamputraintan  Ahh, just found the answer
>when we make the on: pull-request actions, it will automatically use the synchronize event which will always consider all commits of a file change.

So when you make changes out the paths in any commit of this PR, it will always been triggered.
I just have tested on my side, it will works next time when we just have **.md files change in that PR.

https://github.com/actions/runner/issues/2324#issuecomment-1703345084

Comment (raylrui) [reply .github/workflows/prbuild.yml:13]: so actually , we can just add path-ignores this time, test next time 🤣

```
paths-ignore:
 - '**.md'
```

raylrui (approved):
All good for this PR as it include yml file. We can have a bit testing in next PR.
Or you can test with ignore yml file as well....

.github/workflows/prbuild.yml
williamputraintan: Ahhh, so the path filtered is from the PR change, not from the commit. That make sense why it ignores the `path-ignore`. Oh well, will just leave it there in case for docs PR. Thank you for looking in to it!

PR #307, opened 2024-05-18
Added documentation on Event Schema and Namespace guide

victorskl:
* Improved repo landing README.md

PR #308, opened 2024-05-18
Updated WorkflowRunStateChange schema

victorskl:
* Added properties: `workflowName`, `workflowRunName`
* Removed properties: `workflowType`
* Added Payload properties: `data`
* Refactored service namespace: `wfm` to `workflowmanager`


victorskl:
This change request is based on meeting notes 16/05/2024 - https://umccr.slack.com/archives/C03ABJTSN7J/p1715906599365689

victorskl:
Related #307

docs/event-schemas/workflowmanager/WorkflowRunStateChange.json
```diff
@@ -54,8 +54,9 @@
           "portalRunId",
           "timestamp",
           "status",
```
reisingerf, line 56: I wonder if we should rename this to "state"... 
It would align better with the "state change" events.

docs/event-schemas/workflowmanager/WorkflowRunStateChange.json
```diff
@@ -54,8 +54,9 @@
...
-          "workflowType",
+          "workflowName",
           "workflowVersion",
+          "workflowRunName",
           "payload"
         ],
```
reisingerf, line 61: The event is missing the execution id (as defined by the execution service). It's not stricktly needed for the `WorkflowRunStateChange` event from the `orcabus.workflowrunmanager` (if the WRM does not want to communicate it out), but would have to be part of the `WorkflowRunStateChange` event from the execution services.

If we keep those aligned across services, we should add the execution id here. 
If we are fine with a split and every execution service defines it's own WRSC, then we should make this one specific to the WRM, i.e. the payload schema needs to change to make the ref id mandatory (this is only optional if the payload originates from the execution service).

docs/event-schemas/workflowmanager/WorkflowRunStateChange.json
reisingerf: ignore!

victorskl: We discussed together. Conclusion was - state change might constitute multiple combination of different property changes - e.g. `combo(timestamp, status, portalRunId)`

victorskl:
@alexiswl @raylrui Any comments for the PR?

raylrui (approved):
LGTM

docs/event-schemas/workflowmanager/WorkflowRunStateChange.json
victorskl: Just heads up:

We have discussed and thought it through a bit more. Upon discovering the need of "Glue concept" requirement between steps; there gonna be slight changes on "event-schema" WRSC and "data-schema" for Payload contract.

Concepts:
1. Glue
2. Shared WRSC event-schema (for Execution Services) 
3. Dedicated WRSC event-schema (for WorkflowManager) 
4. The "data-schema" contract for Payload structure interface 

Outlooks:
- There will be only 2 WRSC schemas.
- There will be 2 types of schemas - "event-schema" and "data-schema"
- The "data-schema" does not contain (AWS) event envelope properties such as `source`, etc. 
- The "data-schema" are not "event-schema". Just an interface contract. The data-schema can be used as **composition** to form a instance of event message out from "event-schema" such as WRSC.
- Not every workflow execution service needs maintaining WRSC schema. But, only "data-schema" for Payload structure contract.


@raylrui @alexiswl - Florian will reach you out on further explaining these concepts and outlooks. And, we are going to sprint (role play) this, this week. FYI.

victorskl:
Merging it

PR #304, opened 2024-05-17
feat: filemanager inventory

mmalenic:
Closes #219 

### Changes
* Add support for ingesting events using [S3 inventory](https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-inventory.html) (or a custom inventory file uploaded to S3).
   * Events are parsed by fetching the `manifest.json` file and converting to database `FlatS3EventMessage` records.
   * Inventory events do not have sequencer values, so in order to bind to following deleted events, the sequencer is set to the empty string.
* Add integration tests using AWS generated inventory files.
   * Using some macros to include bytes and [`aws-smithy-mocks-experimental`](https://crates.io/crates/aws-smithy-mocks-experimental), which seems to work pretty well, and is more ergonomic than [mockall](https://crates.io/crates/mockall) and `#[double]`.
   
Note, I've removed Parquet and Orc from this PR so that it's less bulky. I'll will include it in the next one.

### Future
* I'd like to potentially try using some orm functionality from [sea-orm](https://github.com/SeaQL/sea-orm). Since sea-orm uses sqlx, we could opt in to using only the features that we want. For example, the [sea-orm-cli](https://github.com/SeaQL/sea-orm/tree/master/sea-orm-cli) `generate entity`, looks really useful, and could be added to `build.rs`. This inspects the database and generates strongly-typed Rust structs and enums for use in queries. Currently, using `FlatS3EventMessages` or other types is a bit bulky, especially in tests, where some files have more than 50% repetitive test data initialization code (although some well-placed macros would help with this). Let me know your thoughts @brainstorm.

Issue #219, opened 2024-04-15
filemanager: sync database to current S3 state


mmalenic:
There should be functionality that allows the filemanager to sync it's database to the current S3 state. It is possible that the database state becomes out of sync, for example, due to events that fail to be ingested, or due to old objects that existed before the filemanager was deployed.

Some ways to accomplish this:
* Use the S3 Inventory.
    * This has the advantage of allowing someone to manually input an inventory report (i.e. call a filemanager API and sync with the given inventory report).
    * Could be configured to receive an inventory from a bucket (e.g. via [notifications](https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-inventory-notification.html)) on a daily/weekly basis?
    * Cannot be run on demand.
* Call S3 list operations on a bucket.
    * Could be run on demand.

PR #310, opened 2024-05-20
feat(filemanager): add Parquet and Orc support to inventory

mmalenic:
Related to #219 and #304

Some left overs from the previous PR.

### Changes
* Add support for Parquet and Orc for the filemanager inventory function


brainstorm (approved):
Woah, didn't expect that ORC support too, kudos!

mmalenic:
> Woah, didn't expect that ORC support too, kudos!
Effort-wise the same, as they both get deserialized into Arrow, just a different reader.

PR #321, opened 2024-05-24
fix: filemanager delete marker

### LINKED ISSUE #316: filemanager: consider `s3:ObjectRemoved:DeleteMarkerCreated` events when versioning is enabled
The filemanager currently classifies `s3:ObjectRemoved:DeleteMarkerCreated` as object removed events, however they should probably be classified as object created events, as this would better fit the database model. This is because `s3:ObjectRemoved:DeleteMarkerCreated` events are emitted when an object with versioning is deleted and replaced with a 'delete marker'. This delete marker acts as a separate object which itself can be deleted (emitting another `s3:ObjectRemoved` event for the delete marker).

This operation better resembles a created event (i.e. the delete marker was created), followed by a removed event (the delete marker was deleted).  Potentially delete markers could be tracked as a separate `isDeleteMarker` column in the database too.

mmalenic:
Closes #316 

### Changes
* Tracks delete markers in the database and considers a delete marker being created and permanently deleted as a single record in the database.
* Removes coalesced columns in the update_reordered queries, because null values should also be reprocessed into separate database entries.
* Adds tests for these fixes.


brainstorm (approved):
Interesting gotcha described in issue #316 and the implementation seems sound, kudos!

PR #329, opened 2024-05-28
Fix workflow manager issues when handling WRSC events

reisingerf:
Fix the following:
- correct call to handler methods
- unmarshaller usage.
- workflow type/name legacy code
- datetime handling
- handle non-existing `Workflow` record when persisting `WorkflowRun` record


reisingerf:
@victorskl we also have to fix the event rule pattern. The current form does not work, we have to remove the quoting, just not sure how to best do that in type script.

victorskl (approved):

PR #334, opened 2024-05-30
update the dynamodb table name for conflict

raylrui:
update the translate service dynamodb table name for bclconvertmanager update, and fix conflicts in build


raylrui:
Merged as this just for change a dynamo db name

PR #341, opened 2024-06-05
Convert api-url to api-domain

alexiswl:
Resolves error where requests expects env var to be url instead of a domain


victorskl (approved):

victorskl:
Merging...

Issue #331, opened 2024-05-29
filemanager: ingest S3 events directly rather than merging them into `Created`/`Deleted` pairs.


mmalenic:
## Issue
The filemanager currently ingests events as they arrive, linking `Created` and `Deleted` events together on insertion. This should probably be changed to avoid linking them on insertion in order to reduce complexity and improve performance.

### Current approach
The filemanager has a table structure that represents both the `Created` and `Deleted` event from S3 in a single record. On insertion, `Deleted` events are linked to existing `Created` records and out-of-order events are re-ordered by updating existing records.

### New implementation
The change can be implemented by having the tables represent one event, with a type column indicating whether it is a `Created` or `Deleted` event. Note that event the de-duplication logic would remain the same - this mostly only affects how the re-ordering logic works.

#### Advantages
* The insertion logic would be much simpler. Currently there is complicated SQL logic to reorder events on insertion which is more difficult to understand and maintain than a simple `insert into ...` query.
* The insertion logic would likely be faster, and constant across records on the same bucket, key and version_id. I think this probably happens because of the `select *` inside the reorder [query](https://github.com/umccr/orcabus/blob/415729f30823190c0405df59f6f44191e3ed448f/lib/workload/stateless/stacks/filemanager/database/queries/ingester/aws/update_reordered_for_created.sql#L63-L86). See some benchmarks on this [branch](https://github.com/umccr/orcabus/blob/feat/filemanager-bench/lib/workload/stateless/stacks/filemanager/filemanager/benches/ingest_bench.rs) (will upload some graphs soon). I'm planning to eventually integrate this branch as filemanager performance tests/benchmarks, which is related to #229.
* There is slightly more information recorded (e.g. the exact order of incoming events), because each event is directly recorded in the database rather than being merged.

#### Disadvantages
* Query logic is slightly more complicated (although less so than the insertion logic), because queries would need to consider the sequencer value in an `order by ...` clause.
* Queries are potentially less performant as re-ordering has to be done when fetching the data (I haven't tested this).
* The domain model of objects is less intuitive. A record in the `s3_object` table no longer represents a single object, but a point when that object was created/deleted/modified. This could be important for grouping objects. E.g. metadata on a group of objects would probably need to sit on one of the points, such as the `Created` record of an object.

### In-between approach
[...]

PR #346, opened 2024-06-06
refactor: filemanager ingest

mmalenic:
Related to #331

### Changes
* Adds a simplified, direct mode of ingestion by changing the table definition to include a `event_type` column.
* There's a bit of duplicate code because I'm keeping both types of ingestion around, although eventually one of these will be removed.


lib/workload/stateless/stacks/filemanager/filemanager/src/database/aws/ingester_paired.rs
```diff
@@ -0,0 +1,1822 @@
+//! This module handles logic associated with event ingestion using event ingestion logic that
+//! pairs Created and Deleted events.
+//!
+
```
mmalenic, line 4: This file is mostly a direct copy from the main branch [`ingester.rs`](https://github.com/umccr/orcabus/blob/main/lib/workload/stateless/stacks/filemanager/filemanager/src/database/aws/ingester.rs).

brainstorm (approved):
Neat re-use of timestamp field and the new "Paired" mode, thanks for evaluating perf as a part of your changeset too!

victorskl (approved):

mmalenic:
Note, I've refactored this slighty to make use of one table for both ingest modes, so that it will be cleaner to update this if one of the modes gets removed.

PR #355, opened 2024-06-17
fix(filemanager): make `HeadObject` optional

### LINKED ISSUE #354: filemanager: `HeadObject` causes error when there are no permissions
The filemanager attempts to grab the `HeadObject` of ingested events in order to update fields like `last_modified`, `storage_class`, `checksum`, etc. If this fails due to a lack of permissions on the bucket, then ingestion fails. Instead, the filemanager should ingest the event even if `HeadObject` fails, as these fields are optional anyway.

mmalenic:
Closes #354 

### Changes
* Makes the `HeadObject` call for ingesting events optional.

### DIFF
--- a/config/constants.ts
+++ b/config/constants.ts
@@ -161,6 +161,8 @@ export const bclconvertInteropQcIcav2EventSource = 'orcabus.bclconvertinteropqc'
 export const bclconvertInteropQcIcav2EventDetailType = 'WorkflowRunStateChange';
 export const bclconvertInteropQcStateMachinePrefix = 'bclconvertInteropQcSfn';
 
+export const fileManagerIngestRoleName = 'orcabus-file-manager-ingest-role';
+
 /*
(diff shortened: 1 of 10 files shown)


brainstorm (approved):

brainstorm:
Ah, good consistency catch... so we now ignore but log because nothing else can be done I guess 👍

mmalenic:
### Update
* Also added permissions for the ingest Lambda function to access pipeline cache buckets

victorskl (approved):
LGTM.

Just heads up. Conceptually, `HeadObject` ops can determine the object status details once it get moved into Glacier Deep Archive tier or, objects from archival bucket.

reisingerf (approved):

mmalenic:
### Another update

* I've added a `NamedLamdbaRole` construct to support creating named roles.
* Add warning for a lack of permissions.
* Fixed CDK nag.

PR #357, opened 2024-06-17
feat: filemanager sea-orm entities

### LINKED ISSUE #314: filemanager: setup sea-orm with entity generation
To ease the use of writing sqlx queries, sea-orm-cli [entity generation](https://www.sea-ql.org/SeaORM/docs/generate-entity/sea-orm-cli/#generating-entity-files) could be used to generate types for the database tables automatically. Since this is a cli tool, it could be run before compiling code, or potentially to integrate it better with `cargo build`, it could be run inside `build.rs`.

This would also help with the querying API in #109 and #110, as there wouldn't be as much of a need to define types for values returned by the database.

mmalenic:
Closes #314

### Changes
* Setups up sea-orm entity generation in a build.rs script, and includes the generated code as part of the file manager.
    * A new `filemanager-build` crate is used for better code organisation.

Leaving this as a draft for now until the phase 1 deployment is complete.


lib/workload/stateless/stacks/filemanager/filemanager-build/src/entities.rs
```diff
@@ -0,0 +1,48 @@
...
+        panic!("command must be generate");
+    }
+
+    let path = out_dir.join("mod.rs");
+    let path = path.to_string_lossy();
+    let generated = quote!(
+        // Auto-generated by the filemanager build script.
+        #[path = #path]
+        pub mod entities;
+    );
```
mmalenic, line 43: This is required to be able to include the module in the `filemanager` crate using `include!(concat!(env!("OUT_DIR"), "/entities.rs"))`. My IDE is able to pick up these modules okay, AFAIK vscode should work too.

lib/workload/stateless/stacks/filemanager/filemanager-build/src/error.rs
```diff
@@ -0,0 +1,95 @@
...
+                return Self::new(
+                    error_kind,
+                    Some(NamedSource::new(loc.file(), source)),
+                    Some(offset),
+                );
+            }
+        }
+
+        Self::new(error_kind, None, None)
+    }
```
mmalenic, line 79: Using miette for nicer compile/build error printing. `#[track_caller]` is useful to print line information about where an error occurred, similar to the Rust compiler.

brainstorm (changes requested):
If I understand the code correctly, `filemanager-build/src/entities.rs` generates ... `entities.rs` that will be `include!`-d on `mod.rs`... for easier source code navigation I would prefix with `gen_` or similar (i.e `gen_entities.rs`), otherwise it gets confusing what codegens what (`entities.rs` codegenerating another `entities.rs` somewhere else?) :-S

If filemanager-build/src/entities.rs is the generated file, I'd put a header that states that: `Autogenerated file, see the other file gen_entities.rs for its source/generation`.

mmalenic:
That's a good call, `filemanager-build/src/entities.rs` is not a generated file, it does the generation step. So I think for clarity, I'll rename that to `gen_entities.rs`. This creates a file called `${OUT_DIR}/entities.rs` :see_no_evil: ... That file should have a comment in it, and could be called `entities.rs`. Or maybe `gen.rs` so that it's clear where it's included that it represents generated code?

brainstorm:
> That file should have a comment in it, and could be called `entities.rs`. Or maybe `gen.rs` so that it's clear where it's included that it represents generated code?

Either way works, just make it clear **which generates and which is generated** ;)

brainstorm (approved):

victorskl (approved):
LGTM!

PR #362, opened 2024-06-19
Feature(metadata-manager): Scheduled Sync from Metadata-Manager

### LINKED ISSUE #359: Scheduled Sync Tracking Sheet to Metadata Manager
Start loading tracking sheet to metadata manager periodically

williamputraintan:
- Update django model to support naming in convention in TrackingSheet
- Add scheduled event to sync daily in PROD
- Loaded current metadata from tracking sheet to dev Db

Resolves #359


victorskl (approved):

lib/workload/stateless/stacks/metadata-manager/app/models/lab/library.py
```diff
@@ -25,7 +25,7 @@ class Phenotype(models.TextChoices):
 
 
 class Quality(models.TextChoices):
-    VERY_POOR = 'very-poor'
+    VERY_POOR = 'very-poor', 'VeryPoor'
```
reisingerf, line 28: We should fix those inconsistencies in the source..... 
I'll make a note... somewhere...

reisingerf (approved):

PR #369, opened 2024-06-24
Fixed ctTSOv2 workflowVersion to 2.5.0

victorskl:


victorskl:
Not so confident about this. But, pls have a look, Alexis.

alexiswl:
Ah good spot!

victorskl:
Hold on, I found this is static config inside `constants.ts` 
Let me double check.

victorskl:
Ok, guess, I will leave it with you for later refactor/realign these config parameter with cttsov2-pipeline-manager stack which leverage these config param from `constants.ts`.

For now, this fix follow as the way it is...

victorskl:
How about this PR @alexiswl? If you have another plan, happy to on-hold or abandon, etc..

alexiswl (approved):
Hmm, yes using contants.ts would be good but also happy to wait and then whatever replaces stacky can use the constants.ts and we just hardcode for now

victorskl:
Sound good. I will keep an eye on this, too. Part of stacky backlog...

PR #371, opened 2024-06-24
Fixed ICAv2 event rule prefix matching

victorskl:
* This has to be in-synced with ICAv2 workflow trigger side convention


alexiswl (approved):
Happy for this hack for now

PR #382, opened 2024-06-28
Promotion for `postgres-manager` to the stateful stack group (from stateless) 

williamputraintan:
`postgres-manager` will have some state which is the secret manager where the microservice will rely on. Moving this to stateful stack group.


williamputraintan:
the stack in dev is destroyed

victorskl (approved):

williamputraintan:
Already deployed and update manually, so the microservice should work as normal.

PR #383, opened 2024-06-30
Fix: STG Deployment

williamputraintan:
- Comment out Prod stage in stateful pipeline, SG lookup will fail when doesn't exist. SG lookup is needed by the pg-manager
- Move SSM Parameter lookup to resolve result on deployment
- Update `context.json`

### DIFF
--- a/cdk.context.json
+++ b/cdk.context.json
@@ -238,5 +238,9 @@
         ]
       }
     ]
+  },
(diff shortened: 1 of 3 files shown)


raylrui (approved):

victorskl (approved):
👌

PR #389, opened 2024-07-05
Released Stateful Pipeline to Production

victorskl:


victorskl:
Note: I have deploy Shared Stack at this point as per instruction.

lib/pipeline/statefulPipelineStack.ts
```diff
@@ -122,24 +122,24 @@ export class StatefulPipelineStack extends cdk.Stack {
 
     // Some stack have dependencies to the 'shared stack' so we need to deploy it first beforehand
```
williamputraintan, line 123: I guess we could remove this comment now

williamputraintan (approved):
LGTM!

lib/pipeline/statefulPipelineStack.ts
victorskl: Similar thought. 
Let keep the comment there for future init another environment/account.
This comment is good.

PR #396, opened 2024-07-08
fix(filemanager): sequencer ordering

### LINKED ISSUE #395: filemanager: order by sequencer
By default, any API list requests should order results by the `s3_object` sequencer value, to ensure that results are returned in the correct order.

mmalenic:
Closes #395 

### Changes
* Order by sequencer in list API queries.
* Add an index on the sequencer to support better sorting of ingested events.
    * I think this effectively does what this piece of code is doing, except it's a native construct to the database: https://github.com/umccr/orcabus/blob/2e4772c22edf982aef7d64fc4b10c685a716433c/lib/workload/stateless/stacks/filemanager/database/queries/ingester/aws/update_reordered_for_created.sql#L64-L90 This is one of the reasons why I simplified the database ingestion, because I think a similar trade off with ingest/query performance can be achieved by just using an index (although I would need to test this to be sure).

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/Cargo.lock
+++ b/lib/workload/stateless/stacks/filemanager/Cargo.lock
@@ -2145,6 +2145,7 @@ dependencies = [
  "mockall_double",
  "orc-rust",
  "parquet",
+ "rand",
  "sea-orm",
  "serde",
  "serde_json",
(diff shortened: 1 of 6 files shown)


brainstorm (approved):
🚀

victorskl (approved):

PR #399, opened 2024-07-09
fix: filemanager e_tag quoting

### LINKED ISSUE #397: filemanager: s3_object e_tag is surrounded by quotes
Sometimes the `s3_object` e_tag is surrounded by double quotes when ingested into the database, which should be fixed.

mmalenic:
Fixes #397 

Turns out that e_tags [should be double quoted](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) to distinguish between two different kinds of e_tags, weak and strong. An e_tag is quoted when received from an SDK `HeadObjectOutput`. However, EventBridge/SQS sends S3 events with unquoted e_tags, resulting in a mismatch when ingesting events.

### Changes
* Check/modify all e_tags to ensure that they are quoted before ingesting events.
* Move the `default_version_id` to `messages.rs`.


brainstorm (approved):
Wow, nice catch!

victorskl (approved):

PR #424, opened 2024-07-26
feat(filemanager): attribute updating API

### LINKED ISSUE #423: filemanager: attribute updating API
The filemanager API should support updating the attributes, as part of enabling attribute linking between services. The API should accept a PATCH request for attributes for a given object record which modifies the attribute column in the database. [JSON patch](https://jsonpatch.com/) is a good approach for this because it has a standardized way to update JSON fields using HTTP PATCH requests.

mmalenic:
Closes #423 

### Changes
* Adds multiple PATCH routes which allow updating the attributes column for a single `s3_object`/`object`, or a collection of records.
    * This uses JSON patch to perform the update by first querying the database for the correct records to update, then applying the patch to the attributes column, and writing the attributes back to the database.
* Add tests for the new endpoints and update the openapi docs.
* Return `NOT_FOUND` for getting/updating single records if the supplied `id` does not exist.
* Bump a few dependencies.

### To do

This is part of the change that allows attribute linking. In theory, any service can now add whatever attributes they want to a filemanager record. Still to do is implementing a rule system to submit jobs for linking attributes on filemanager records, and allowing attributes to be updated using the EventBus instead of a PATCH request.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/Cargo.lock
+++ b/lib/workload/stateless/stacks/filemanager/Cargo.lock
@@ -2137,6 +2137,7 @@ dependencies = [
  "futures",
  "hex",
  "itertools 0.13.0",
+ "json-patch",
  "lambda_runtime",
  "lazy_static",
  "md5",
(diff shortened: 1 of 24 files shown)


victorskl (approved):

mmalenic:
### Update
* Fixed an error when running the local `filemanager-api-server` due to a missing "env-filter" feature flag on "tracing-subscriber".
* Added ability to specify current state for the update s3_object attributes endpoint

PR #425, opened 2024-07-26
Feat: Running Microservice API Locally

williamputraintan:
- Added to run Microservices API to run locally simultaneously via docker (except filemanager)
- Added to load SQL dump from S3 (data is dumped from dev cluster) to local server

Todo:
- add filemanager service to the docker-compose (followed up by @mmalenic )

PR #429, opened 2024-07-31
fix: add filemanager to local API setup

mmalenic:
Following up from #425

### Changes
* Add Dockerfile for building filemanager.
* Add filemanager to the local microservices setup from the top-level directory.
    * This loads the most recent 100000 records from the dev database. 
* Fix env variable parsing for `filemanager-api-server`.

This was a bit tougher than I expected. One of the problems is that the filemanager relies on having access to a database to generate code for types/schemas at compile time. This is challenging in a Docker context, because the build process is not run in the compose network, and there is no straightforward way to have a database running before building the filemanager Dockerfile.

In the future, I think I might remove the compile-time database requirement from filemanager. It seems like it often gets in the way, and tools seem to expect that no database is required to compile/build code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,10 @@ You can run all microservice APIs locally at once using Docker Compose provided 
 make start-all-service
 ```
 
+This command loads SQL data dumps from S3 in the dev account, and assumes that the shell is logged in to AWS. The SQL
+dumps generally contain all records from the dev database, except for filemanager, which has 100000 of the most recent
+records (due to it's size).
+
 To stop the services, use:
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -130,10 +130,10 @@ The APIs will run on `localhost` with the following port assignments:
 
 | Microservice         | Local Endpoint          |
 |----------------------|-------------------------|
-| File Manager         | <http://localhost:8000> |
 | Metadata Manager     | <http://localhost:8100> |
 | Workflow Manager     | <http://localhost:8200> |
 | Sequence Run Manager | <http://localhost:8300> |
+| File Manager         | <http://localhost:8400> |
 
 ### Typography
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -116,6 +116,10 @@ You can run all microservice APIs locally at once using Docker Compose provided
 make start-all-service
 ```
 
+This command loads SQL data dumps from S3 in the dev account, and assumes that the shell is logged in to AWS. The SQL
+dumps generally contain all records from the dev database, except for filemanager, which has 100000 of the most recent
+records (due to it's size).
+
 To stop the services, use:
 
 ```sh
```
williamputraintan, line 125: Perhaps if you could change this readme that filemanager is using port `8400` instead of `8000` would be great 😊

williamputraintan (approved):
Spin this locally and LGTM! Thanks for this follow up PR!

victorskl (approved):

README.md
mmalenic: Good catch!

brainstorm (approved):
I understand the problem w.r.t docker and sqlx compile time validation... It's really good for local development but not for CI. I wonder how other devs using this tooling cope with it.. probably via django-like env vars and conditionally disabling the compile time checks instead of turning them off for good.

mmalenic:
Yeah it depends. I'm eventually planning on transitioning fully to sea-orm, which doesn't have compile-time checks in the same way. It does have code generation for entities though, and it seems like others just check these into the code base, e.g. see https://github.com/SeaQL/sea-orm/blob/master/COMMUNITY.md#built-with-seaorm.

For sqlx it's difficult to conditionally turn things on/off because the compile-time checks either happen when using macros, or not. So you can either write functions with compile-time checks using the `query!(...)` macros, or without compile-time checks using the `query(...)` functions. I think this is because the `query!` macros need to generate struct definitions from the database schemas.

There is always the offline mode using `sqlx prepare`, but that wasn't the nicest in terms of checking in auto-generated code. :smiling_face_with_tear:

Issue #431, opened 2024-07-31
Use Parallel Boxes in AWS Step Functions to reduce output sizes


alexiswl:
See https://ap-southeast-2.console.aws.amazon.com/states/home?region=ap-southeast-2#/v2/executions/details/arn:aws:states:ap-southeast-2:843407916570:execution:wtsSfn-wfm-ready-event-handler:6bf934b6-f261-fc9e-70a9-07397476960f_4462aed6-bc08-5dbb-fb81-d9154dfdd653 as an example.

```
{
  "cause": "The state/task 'Update DataBase Post Launch' returned a result with a size exceeding the maximum number of bytes service limit.",
  "error": "States.DataLimitExceeded"
}
```

Write to databases where possible and then wrap the step and the write to database step in one parallel box

PR #432, opened 2024-07-31
Wrap launch analysis step function segment into a parallel chunk to reduce step output sizes

alexiswl:
Resolves https://github.com/umccr/orcabus/issues/431


alexiswl:
Ah sorry @victorskl merged via habit

victorskl:
All good.

PR #433, opened 2024-07-31
Fixed cttsov2 post-processing docker architecture ARM_64

victorskl:
skip-checks:true


victorskl:
Skipping CI as it is quick build fix. It is going to run again in CodePipeline.

alexiswl:
Thanks for this! Sorry, I don't think there's any other ones. Because the docker image compiles tabix it's extremely slow to do a cross-compilation

PR #451, opened 2024-08-06
Add tags to event data for workflowrunstatechange events

alexiswl:
Use tags, useful for filtering events.

Remove cttsov2 layers, each lambda can run their own set icav2 env vars functions.  Layer size was exceeding limit

Fill placeholders now requires ICAv2 credential access since we convert icav2 uris to s3 uris for events

Use s3 for output uris for all icav2 services


victorskl (approved):

reisingerf (approved):
Let's try it out

victorskl:
Merging this for now, Alexi. We can follow up another PR, if any.

PR #457, opened 2024-08-08
feat: filemanager API wildcards

### LINKED ISSUE #456: filemanager: improve API to include wildcard matching
The filemanager API could be more expressive if it includes wildcard to support matching objects with certain prefixes/suffixes. The wildcards could be used for list queries, and for update queries. Wildcards would also be useful in attributes to support querying/updating attributes for a collection of records.

mmalenic:
Closes #456

### Changes
* Add wildcard matching to API routes for list/update.
    * Wildcards use '%' for multiple characters and '_' for single characters, which get converted to `like` queries.
    * `case_sensitive` can be used to control whether the query is `like` or `ilike`.
    * Wildcards are also supported in JSON attributes.
* Add API guide and design docs.
* Update tests and OpenAPI definitions.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/README.md
+++ b/lib/workload/stateless/stacks/filemanager/README.md
@@ -86,7 +86,7 @@ Alternatively, just `brew install dbeaver-community` to easily browse the databa
 
 ## Local API server
 
-To use the local API server, run:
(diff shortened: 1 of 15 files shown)


victorskl (approved):
LGTM. Thanks for note taking around API and attribute linking.

PR #465, opened 2024-08-12
refactor(filemanager): remove object table

### LINKED ISSUE #464: filemanager: remove `object` table
The object table doesn't serve much of a purpose anymore and should be removed, as we decided to use the `attributes` column on `s3_object` to store any attributes.

If there is a requirement to have attributes follow relational modelling more closely (instead of JSON), or to store non-S3 data, then another table can be added back in. This would probably look slightly different to the current `objects` table anyway as it wouldn't be a 1-many relationship.

mmalenic:
Closes #464 

### Changes
* Remove the `object` table.
* Also remove `public_id` from `s3_object` as it isn't being used. This can be reintroduced if necessary.
* Remove some tests as they are more related to custom re-ordering logic using `paired_ingest_mode`, rather than the default simpler ingestion.
* Rename routes/functions.


victorskl (approved):
LGTM

PR #467, opened 2024-08-12
fix(filemanager): local s3-load should not use `object`

mmalenic:
Related to #464

### Changes
* Fix `Makefile` so that it does not use `object` data dumps when loading from s3.
    * Data dump has been updated on S3.


victorskl (approved):
Thanks!

PR #472, opened 2024-08-13
refactor(filemanager): API style

mmalenic:
Closes #458

### Changes
* All responses/requests are `camelCase`.
* Routes are shortened to not use `-` and `_`.
    * Swagger ui uses `swagger-ui` instead of `swagger_ui`.
* Pagination uses links and contains count.

PR #475, opened 2024-08-15
feat: filemanager presigned route

### LINKED ISSUE #459: filemanager: add presigned URL route
The orcabus UI needs access to presigned URLs for fetching objects. A new route could be added for a single object/multiple objects by key:

* `/api/v1/s3_objects/presign/{:id}`
* `/api/v1/s3_objects/presign?key=...`.
    * This could support any of the query filter parameters.
    * Should it also support pagination?

This could return a response with a single URL (and potentially a redirect?) or multiple URLs in the case of a collection of objects:
```json
{
    "results": [ "<presigned_urls...>" ]
}
```

mmalenic:
Closes #459

### Changes
* Adds routes under `/s3/presign` and `/s3/presign/{:id}` to fetch presigned URLs for current objects only.
* `responseContentDisposition` can be used to set the content-disposition of the presigned `GetObject` request. The default value is `inline`, as this matches the HTTP `content-disposition` header.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md
+++ b/lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md
@@ -159,6 +159,30 @@ For example, count the total records:
 curl -H "Authorization: Bearer $TOKEN" "https://file.dev.umccr.org/api/v1/s3/count" | jq
 ```
 
+## Presigned URLs
+
+The filemanager API can also generate presigned URLs. Presigned URLs can only be generated for objects that currently
+exist in S3.
+
+For example, generate a presigned URL for a single record:
+
+```sh
+curl -H "Authorization: Bearer $TOKEN" "https://file.dev.umccr.org/api/v1/s3/presign/0190465f-68fa-76e4-9c36-12bdf1a1571d" | jq
+```
+
+Or, for multiple records, which supports the same query parameters as list operations (except `currentState` as that is implied):
+
+```sh
+curl -H "Authorization: Bearer $TOKEN" "https://file.dev.umccr.org/api/v1/s3/presign?page=10&rowsPerPage=50" | jq
+```
+
+Specify `responseContentDisposition` for either of the above routes to change the `response-content-disposition` for the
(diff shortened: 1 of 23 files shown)


victorskl (approved):
LGTM

brainstorm (approved):

mmalenic:
### Update
* Fixed a bug with parsing the `HOST` header, which was introduced with #472. Thanks @williamputraintan.
    * I assumed that `HOST` was a valid URL, but it's actually just the `host:port` without the scheme.
    
* I've added a `FILEMANAGER_API_TLS_LINKS` option which controls whether links are returned as `http` or `https`.
    * Should this be more automatically derived? E.g. default `http` for `filemanager-api-server`, and default `https` for `filemanager-api-lambda`?

williamputraintan (approved):

williamputraintan (approved):

williamputraintan:
While this is not merged, possible to add an env-var in the compose for the filemanager service? So the local API server could have some presigned URL instead of an `unhandled error`?

```
      # - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-random_access_key_id}
      # - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-random_secret_access_key}
      # - AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-ap-southeast-2}
```

lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md
```diff
@@ -150,6 +160,30 @@ For example, count the total records:
...
 
+## Presigned URLs
+
+The filemanager API can also generate presigned URLs. Presigned URLs can only be generated for objects that currently
+exist in S3.
+
+For example, generate a presigned URL for a single record:
+
+```sh
+curl -H "Authorization: Bearer $TOKEN" "https://file.dev.umccr.org/api/v1/s3/presign/0190465f-68fa-76e4-9c36-12bdf1a1571d" | jq
```
williamputraintan, line 171: A quick question, the API only support presigned from an `object-id`?

PR #476, opened 2024-08-15
[skip ci] Added dependabot.yml config file

victorskl:
* Append commit message prefix with skip ci initially.


victorskl:
Look like `[skip ci]` skip check all together. I will try with different tag `[draft]`.

PR #477, opened 2024-08-15

victorskl:
* Append commit message prefix with skip ci initially.


victorskl:
Ok, I will split `prbuild.yml` and update that one first. 
Then, I will test whether `[draft]` tag work.

PR #479, opened 2024-08-16
[skip-ci] Updated GH prbuild.yml

victorskl:
* Check also PR title contain `[draft]` tag


victorskl:
My bad. Correct [GH syntax](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs) is no dash. `[skip ci]`

I will try again.

Issue #455, opened 2024-08-08
Configure dependabot PR to draft mode initially


victorskl:
**Context:**

Configure dependabot PR to draft mode initially. So that it won't kick off GH action CI PR build just yet (which consider expensive ops if dev/team decide to abandon PR later for another good reason e.g. sweep bumps, breaking changes check, etc..). 

**Approaches:**

- See https://github.com/dependabot/dependabot-core/issues/1291
- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

victorskl:
GH expr syntax:

- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions

victorskl:
Further config pointers:

- https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

PR #480, opened 2024-08-16
[skip ci]: Updated GH prbuild.yml

victorskl:
* Check also PR title contain `[draft]` tag


victorskl:
Changes contains in PR is precondition for #455
So that we can treat dependabot PR to tag `[draft]`. Same as PR draft mode. Won't run CI initially.

PR #482, opened 2024-08-16

victorskl:
* Append commit message prefix with draft mode initially. We shall review PR then figure whether proceed trigger running ci build or skip otherwise.


victorskl:
Ok. Creating PR with Draft Mode is still working. It skips PR builds initially. Does colon `:` matter? Let me try that next.

PR #485, opened 2024-08-16

victorskl:
* Append commit message prefix with [draft] tag initially.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# See https://github.com/umccr/orcabus/issues/455
+
+version: 2
+updates:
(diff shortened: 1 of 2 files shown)


victorskl:
Nice.!

At this point, PR is created with pending CI test / PR build check.

If we find that PR is to be merged, we can resume CI test / PR build by modifying PR title removing `[draft]:` tag. Or, trigger them through GH Action UI > re-run all jobs, etc...

If PR is not to be merged after review then no build need run.

victorskl:
Ok. I will merge this setting file for now. And, I will observe/test with next coming dependabot PR.

Issue #462, opened 2024-08-09
filemanager: consider making the attribute API more user-friendly


mmalenic:
The attribute API uses square brackets for access, which is less ideal because they have to be percent-encoded in the URL. It might be better allow top-level JSON fields to just be accessed directly from the API, instead of nesting them.

E.g, instead of:
`/api/v1/s3_objects?attributes[portal_run_id]=...`

It could be:
`/api/v1/s3_objects?portal_run_id=...`

Similarly, the JSON patch request does not need to have the outer `attributes` field, as it's clear that only attributes can be patched.

E.g. instead of:
```json
{
    "attributes": [ { "op": "add", "path": "/attribute_id", "value": "attribute_id" } ]
}
```

It could just be:
```json
[ { "op": "add", "path": "/attribute_id", "value": "attribute_id" } ]
```

PR #492, opened 2024-08-16
refactor(filemanager): misc simpler and clearer API

mmalenic:
Some more misc changed, related to #462

### Changes
* JSON patch now supports append only, so "add", "test", or "copy" only if an existing key is not being overwritten.
* The outer "attributes" on patch requests is not required.
* `date` has been renamed to `eventTime` for more clarity. /cc @williamputraintan 

I'm holding off on implementing the shortened attributes query string feature, e.g. `portal_run_id=...` rather than `attributes[portal_run_id]=...` as described in #462. This is slightly more complex as it involves capturing unknown parameters in the API. I'm not sure if this is the best idea, because it has the potential to conflict with existing parameters.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/database/migrations/0005_rename_date.sql
+++ b/lib/workload/stateless/stacks/filemanager/database/migrations/0005_rename_date.sql
@@ -1,0 +1,2 @@
+-- Rename `date` to `event_time` as it is clearer what the meaning is compared to `last_modified_date`.
+alter table s3_object rename column date to event_time;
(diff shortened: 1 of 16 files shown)


victorskl (approved):
LGTM. 

Noted well about the challenge with the query parameter impl. We can brainstorm together more.

brainstorm (approved):
Makes sense to not drop `attributes[]` prefix, I suspected it would clash with some other namespace and it seems it indeed does? :-/

PR #501, opened 2024-08-19
Removed dependabot.yml config and revert to its default setting

victorskl:
* It is not straight forward scope only to security-updates. Monorepo
  with nested toolchains get tricky. Reverting back to default.
* Bumped dependencies


victorskl:
Close also #494 #496 #497 #498

Issue #470, opened 2024-08-12
metadata-manager: Id refactoring


williamputraintan:
- The `id` (in each table) -> `orcabusId` (with new Id convention prefix followed by [ulid](https://github.com/ulid/spec) e.g. `lib.XXXXXX`)
- InternalId renaming:
	- library `internal_id` -> ~`lab_id`~ `libraryId` (we will retain this naming)
	- specimen `internal_id` -> `specimenId`
	- subject `internal_id` -> `subjectId`

alexiswl:
Can you let know when you have a PR for this, a lot of step functions will need to be updated for this change. 

Would it be possible to provide both internal_id and lab_id for some time and then remove internal_id?

williamputraintan:
Yes, happy to have at least support the `internal_id` translator. I suppose these `internal_id` names are somehow confusing.

PR #503, opened 2024-08-20
Add support for linked libraries to WRSC event schema

reisingerf:
This is a new convention to support the linking of workflow runs to libraries.

The association currently has to be supported by the execution services / stacky glues that have that information, as the workflow manager does not have access to it atm.
Also, the workflow manager currently relies on the WRSC event to establish the library association, including the `libraryId` (from the lab metadata) and the corresponding `orcabusId` (from the metadata manager). This is expected to be relaxed once we have a different mechanism of updating the workflow manager (e.g. `LibraryStateChange` events).

@raylrui @williamputraintan 

For now this addition is optional as to not break any existing implementations. However, it would be great if we could adopt this wherever possible.

### DIFF
--- a/docs/schemas/events/executionservice/WorkflowRunStateChange.schema.json
+++ b/docs/schemas/events/executionservice/WorkflowRunStateChange.schema.json
@@ -77,8 +77,29 @@
         "workflowRunName": {
           "type": "string"
         },
+        "linkedLibraries": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/LibraryRecord"
+          }
+        },
         "payload": {
           "$ref": "#/definitions/Payload"
+        }
+      }
+    },
+    "LibraryRecord": {
+      "type": "object",
+      "required": [
+        "libraryId",
+        "orcabusId"
+      ],
+      "properties": {
+        "libraryId": {
+          "type": "string"
+        },
+        "orcabusId": {
+          "type": "string"
         }
       }
     },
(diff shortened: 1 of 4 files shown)


victorskl (approved):
LGTM

alexiswl:
Looks great! 

Where can we link the orcabus id and corresponding library id?

Currently I get a 404 on the development library API endpoint?

```
export ORCABUS_TOKEN="$(  \
  aws secretsmanager get-secret-value \
    --secret-id orcabus/token-service-jwt \
    --query 'SecretString' \
    --output text | \
    jq --raw-output '.id_token' \
)"
```

```
curl --silent --show-error --location --fail-with-body \
 --header "Authorization: Bearer $ORCABUS_TOKEN" \
 --url "https://metadata.dev.umccr.org/library?rowsPerPage=1000"
```

Gives

```
<!doctype html>
<html lang="en">
<head>
  <title>Not Found</title>
</head>
<body>
  <h1>Not Found</h1><p>The requested resource was not found on this server.</p>
</body>
</html>
curl: (22) The requested URL returned error: 404
```

reisingerf:
Due to the change to include `/api/v1/` to the endpoint path.

So use `https://metadata.dev.umccr.org/api/v1/library`

reisingerf:
The addition of the `orcabus_id` is still pending I guess. We just introduced this...

victorskl:
> Where can we link the orcabus id and corresponding library id?

We get it from Metadata Manager, yes.

@williamputraintan is working on it as part of id refactor ticket #470

alexiswl (approved):

PR #505, opened 2024-08-20
Stacky updates for generating umccrise events

alexiswl:
Added the stateful umccrise stacky database
Added the logic parts for generating a umccrise ready event

Easier to merge this now knowing at least the umccrise stateful db will be created and there won't be conflicts if I redeploy locally.


victorskl (approved):
Sound good.!

PR #507, opened 2024-08-20
Fix/filemanager api tidy

### LINKED ISSUE #491: filemanager: API and Pagination Request
- `count` in pagination does not return the total record counts but is the length of the results array. I think it should instead return the total record counts for pagination purposes.

- Allow `OPTIONS` method for API. I think this causes cors-error when calling API to the `filemanager`

- Perhaps `page` in pagination should start at 1 instead of 0 (and 0 should be an invalid number). I think normally pages start from 1.

- Presign list swagger doesn't reflect the results.

mmalenic:
Closes #491 

### Changes
* Pagination starts at 1 now, and the count field contains the real count of records for a collection.
* The API server has proper JSON-formatted error responses now, and the OpenAPI presign route is fixed.
* CORS support has been added, which can be adjusted with `FILEMANAGER_API_CORS_ALLOW_ORIGINS`.
    * I've also moved the `corsAllowOrigins` to the top-level config for all microservices.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -70,6 +70,7 @@ services:
       # Container database address for running server inside a docker container.
       - DATABASE_URL=postgresql://orcabus:orcabus@db:5432/filemanager
       - RUST_LOG=debug
+      - FILEMANAGER_API_CORS_ALLOW_ORIGINS=${FILEMANAGER_API_CORS_ALLOW_ORIGINS:-http://localhost:8400}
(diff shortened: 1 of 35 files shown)


victorskl (approved):
LGTM

williamputraintan (approved):
LGTM

williamputraintan:
Ahh, I think the swagger ui seems to have some error with the OpenAPI schema?

Screenshot 2024-08-20 at 17 58 39

mmalenic:
Thanks!

PR #508, opened 2024-08-20
metadata-manager: model refactoring

williamputraintan:
- Refactor `internal_id` to `library_id`, `subject_id`, or `specimen_id`
- Introduce `orcabus_id` as primary key
- Add `ProjectOwner` and `ProjectName` from the tracking sheet to the Library model
- Update OpenAPI docs to document query by field values

Resolve #470 
Resolve #446

PR #509, opened 2024-08-20
Updated stacky with new metadata manager

alexiswl:
Waiting on #508

### DIFF
--- a/lib/workload/components/python-metadata-tools-layer/metadata_tools_layer/src/metadata_tools/utils/library_helpers.py
+++ b/lib/workload/components/python-metadata-tools-layer/metadata_tools_layer/src/metadata_tools/utils/library_helpers.py
@@ -16,7 +16,7 @@ def get_library_from_library_id(library_id: Union[int | str]) -> Dict:
     if isinstance(library_id, str):
         # We have an internal id, convert to int
         params = {
-            "internal_id": library_id
+            "library_id": library_id
         }
     else:
         endpoint = f"{endpoint}/{library_id}"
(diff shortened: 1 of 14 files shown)


lib/workload/components/python-metadata-tools-layer/metadata_tools_layer/src/metadata_tools/utils/library_helpers.py
```diff
@@ -94,6 +94,28 @@ def get_library_workflow(library_id: str) -> Union[str | None]:
...
 
+def get_library_project_owner(library_id: str) -> Union[str | None]:
+    """
+    Given a library id, collect the library project owner
+    :param library_id:
+    :return:
+    """
+    library = get_library_from_library_id(library_id)
+
+    return library.get("project_owner")
```
williamputraintan, line 105: On the casing end, I suppose this `get` will take value directly from the results JSON?  (From #502) the metadata now has a middleware to translate snake_case to pascalCase aligning with other microservices, so this might not get project_owner properly.

In terms of query params, the API may still support snake_case filters, but would recommend using the pascalCase if you can as this is what the Swagger documents it.

alexiswl: I guess this is the same for library_id/Id and subject_id/Id too?

williamputraintan: Yes, the response will be in camelCase

williamputraintan: Yes, the response will be in camelCase

victorskl:
@alexiswl Perhaps, does this PR outdated by #525?

PR #514, opened 2024-08-23
refactor: filemanager indexes

### LINKED ISSUE #460: filemanager: investigate slow queries
Some of the filemanager API routes can be quite slow, such as attribute-based queries, or queries involving the current state. These may need some indexes/partitions to improve speeds, or a rethink on the database table/query design.

mmalenic:
Closes #460

### Changes
* Rework JSON queries to use jsonpath operators and a gin index.
    * The wildcard operators have changed to `*` for multiple characters and `?` for single characters. \cc @williamputraintan 
* Add index for prefix searching on keys.

### To do
* Partition records into `current` objects, and `historical` records.


victorskl (approved):
Sound good 👍

PR #515, opened 2024-08-23
filemanager: attributes route and misc CORS changes

### LINKED ISSUE #495: filemanager: add way to shorten attribute API querying
As described in #462, ideally there would be a way to shorten querying top-level `attributes`:

E.g, instead of:
`/api/v1/s3_objects?attributes[portal_run_id]=...`

It could be:
`/api/v1/s3_objects?portal_run_id=...`

This is a bit of a challenge, as described in #492, because it involves capturing unknown parameters in the API.

This may not be the best idea, because it could lead to confusion and conflicts between the known parameters and unknown parameters. E.g. if the user typed `?object_id=...` it might be better to reject this request rather than assuming that this request was for the `attributes`. Note, that currently unknown parameters are not rejected with an error status code, which should probably be changed.

Some ideas on how to implement this:
* Allow any unknown parameters and assume that the user intends to query by `attributes`.
* Add a list of "allowed" parameters which can be used to query `attributes`. For example, just allowing `portal_run_id` and other important `attributes`. These could be configurable during deployment.

mmalenic:
Closes #495
Depends on #460 

### Changes
* Add a new route under `/api/v1/s3/attributes` which accepts arbitrary parameters that can be a shortcut to querying by top-level attribute properties.

### Misc
* Make the CORS allow methods and allow headers configurable.
* Mirror the `content-type` and `content-encoding` from the client when creating presigned URLs.


victorskl:
Reviewing...

compose.yml
```diff
@@ -71,6 +71,7 @@ services:
       - DATABASE_URL=postgresql://orcabus:orcabus@db:5432/filemanager
       - RUST_LOG=debug
       - FILEMANAGER_API_CORS_ALLOW_ORIGINS=${FILEMANAGER_API_CORS_ALLOW_ORIGINS:-http://localhost:8400}
```
williamputraintan, line 73: Perhaps this could be `http://localhost:3000`, as we set the local port for UI to 3000.

williamputraintan (approved):
LGTM! Thanks for the fix!

williamputraintan:
As a side note when the UI is deployed, we still need to adjust the cors from the s3 bucket side 😅

victorskl (approved):

reisingerf (approved):

mmalenic:
> As a side note when the UI is deployed, we still need to adjust the cors from the s3 bucket side 😅

This would be outside of the OrcaBus stack right, as the bucket configuration is not controlled in this repo?

victorskl:
Yup. 

Will; please feel free to PR for required/desired CORS setting at bucket TF.
You can take ILMN CORS setting as starter - around [these lines](https://github.com/umccr/infrastructure/blob/197f23e7bd8f370697c6bda0ce0a343af8ed712f/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf#L170).

victorskl:
Good to merge?

PR #516, opened 2024-08-24
Fixed WorkflowRun State creation business logic

victorskl:
* WorkflowRun State creation is WRSC timestamp dependant.
  It needs to check "time window" condition before saving
  and emitting (relaying) WRSC event.


reisingerf (approved):
OK with this.
Will change with the next refactor anyway

victorskl:
Sure, understood. I will merge it for now.

PR #521, opened 2024-08-28
Fix/workflowrun state handling

reisingerf:
FYI: @alexiswl 

This should fix the WRSC handling of linked libraries, etc....


victorskl (approved):
Nice refactor.!

alexiswl:
Ah I should move all of my statuses to uppercases then

alexiswl:
Should non-wrsc statuses also change to upper case? 

i.e should `SubjectInSamplesheet` become `SUBJECT_IN_SAMPLESHEET`?

reisingerf:
I think it would be good to follow some sort of convention, yes.

For now the workflow manager changes everything to upper case and replace "-" to "_".

But in your case, these are not WRSC statuses, but events internal to the stacky logic, no?

alexiswl:
> But in your case, these are not WRSC statuses, but events internal to the stacky logic, no?

Yes these are internal events not going through the workflow manager

PR #522, opened 2024-08-29
Fix issue with WRSC generation when a libary record exists

reisingerf:
- Update Django query to expect and return a single DB record instead of a QuerySet
- Handle DoesNotExist case
- Add new test case to check for this condition


victorskl (approved):
Nice catch, Flo.!

PR #524, opened 2024-08-30
WFM - Extend API capabilities - ongoing workflowrun filter

reisingerf:
- Add ongoing endpoint to allow listing of non-terminal workflow runs
- Update mock data to include non-terminal workflow run states


victorskl (approved):
LGTM

PR #525, opened 2024-09-01
Feature/add umccrise rnasum and glues

alexiswl:
## Updates

* Use portal run id for database uuid in icav2 services (since refId is no longer generated if it doesnt exist)
* Use portal run id as idempotency key, allows for lambda retries if the workflow launch fails
* Use upper case (screaming case) for statuses
* Added linkedLibraries for WorkflowRunStateChange payloads
* Added consistency for stacky rule names (will be useful for replay events, particularly if the FilterArns allows for the asterisk on suffixes)
* Added project level events for when fastq list rows are available.  
* Added project name and owner to library event objects
* Updated fastq list row event schema such that rgid is now the fastq list row id. 
* Added arriba output to the wts pipeline (required input for rnasum)


victorskl:
Review in tick, next

reisingerf (approved):
LGTM
(didn't go through all the details though)

Comment / question for upcoming releases: 
There are a few constants that are "hard coded" or SSM parameter controlled, such as pipeline versions, ref files, ... 
We could think about whether that info should come from the system itself, e.g. the workflow IDs could come from the workflow manager and/or the journey service, ref data is that a constant or dependent on the workflow / analysis use case (and could be annotated there instead)?

alexiswl:
> There are a few constants that are "hard coded" or SSM parameter controlled, such as pipeline versions, ref files, ...
We could think about whether that info should come from the system itself, e.g. the workflow IDs could come from the workflow manager and/or the journey service, ref data is that a constant or dependent on the workflow / analysis use case (and could be annotated there instead)?

Yes certainly, I think a lot of these hard-coded parameters would be better off in a terraform file or equivalent

victorskl (approved):
LGTM

PR #529, opened 2024-09-03
filemanager: annotate portal run id

### LINKED ISSUE #528: Add microservice which updates filemanager records
A new microservice should be created which can update records on the filemanager database with the `portalRunId`. The microservice should listen to `WorkflowRunStateChange` events in order to obtain the `portalRunId` and query the filemanager API to update records.

mmalenic:
Closes #528 

### Changes
* Add a new microservice which annotates filemanager records with the portal run id.
    * This consumes `WorkflowRunStateChange` events, for `SUCCEEDED`, `FAILED`, and `ABORTED` states, and calls the filemanager API with a `PATCH` request for updating the attributes using a `/api/v1/s3?key=*<portal_run_id>*` query.
* This is a single Lambda function which is deployed using [`GoFunction`](https://docs.aws.amazon.com/cdk/api/v2/docs/@aws-cdk_aws-lambda-go-alpha.GoFunction.html), with the filemanager endpoint, and reference to the secret containing the JWT token.
* Integration tests with the filemanager microservice are implemented using [testcontainers](https://golang.testcontainers.org/).
    * There's a slight fix of the filemanager `compose.yml` as it was not possible to launch the original compose file due to the build requiring access to a postgres database. This whole setup would be simplified if filemanager didn't need access to a development database to compile code.


lib/workload/stateless/stacks/fmannotator/handlers.go
```diff
@@ -0,0 +1,45 @@
...
+		return err
+	}
+
+	req, err := NewApiClient(config, bytes.NewBuffer(patch))
+	if err != nil {
+		return err
+	}
+
+	req = req.WithMethod("PATCH").WithS3Endpoint().WithQuery(url.Values{
+		"key": {fmt.Sprintf("*%v*", event.Detail.PortalRunId)},
```
reisingerf, line 31: Could add the `/` before & after the portalRunId to make sure we are hitting the intended path part and not some random filename / string in some other part of the key.
But I guess the chances of that happening are negligible, especially since we are not using the portalRunId for anything else in the key....

reisingerf (approved):
LGTM

lib/workload/stateless/stacks/fmannotator/handlers.go
mmalenic: Good idea. 

As an additional comment, at some point it might be nice for the `WorkflowRunStateChange` event to also contain an `outputPath` where objects are expected to turn up. This would allow this service to query the filemanager API with `/api/v1/s3?key=<output_path>/<portal_run_id>/*`, which should be faster because it will hit the index that supports prefix matching, and it would probably be even more robust.

victorskl (approved):

mmalenic:
### Recent updates
* Getting CodeBuild to run and compile filemanager using testcontainers. Related: https://github.com/testcontainers/testcontainers-go/issues/2767.

lib/workload/stateless/stacks/fmannotator/handlers.go
reisingerf: Hm,... 

To some extend that's already in place... WorkflowRunStateChange events can contain a `Payload` JSON which may contain that information. However, this is not enforced nor standardised. That's on purpose (at least for now) to keep the use cases flexible. We may change that in the future... 

Also, I am not sure that would always work... In the normal case the portalRunId update would happen immediately after the objects have been written so we would not expect a key change. However, especially the prefix of the key can change (e.g. hot storage vs archive) and we try to focus on the key part starting from the portalRunId.
So depending on when this procedure is run, this may or may not work.

mmalenic: Okay makes sense. I think we leave it for now and see if it emerges later as a possibility.

lib/workload/stateless/stacks/fmannotator/api.go
```diff
@@ -0,0 +1,94 @@
...
+// The ApiClient which is used to send requests to filemanager.
+type ApiClient struct {
+	Client  *http.Client
+	Request *http.Request
+}
+
+// NewApiClient Create a new ApiClient.
+func NewApiClient(config *Config, body io.Reader) (*ApiClient, error) {
+	req, err := http.NewRequest("GET", config.FileManagerEndpoint, body)
+	if err != nil {
```
brainstorm, line 43: No way to have those as a simple macro or similar in Go? Tons of 3 line repetition :-S

brainstorm (approved):
LGTM other than a bit too repetitive error control ;)

lib/workload/stateless/stacks/fmannotator/api.go
mmalenic: Other than putting the `if` condition on the same line:

```golang
req, err := http.NewRequest("GET", config.FileManagerEndpoint, body); if err != nil {
    return err
}
```

The vibe online seems to be to just embrace it :)

mmalenic: "Luckily, Go does not support macros." - https://stackoverflow.com/questions/49527584/how-to-implement-macros-in-go

PR #531, opened 2024-09-03
Feat: Update API Gateway Configuration

williamputraintan:
- Group the APIGW microservice config as one props across microservice.
- Integrate [orcaui's cognito](https://github.com/umccr/infrastructure/pull/472) to microservice APIGW
- Update some CORS config on APIGW


lib/workload/stateless/stacks/filemanager/deploy/stack.ts
```diff
@@ -141,20 +137,21 @@ export class Filemanager extends Stack {
...
 
     const apiIntegration = new HttpLambdaIntegration('ApiIntegration', apiLambda.function);
 
-    new HttpRoute(this, 'HttpRoute', {
+    new HttpRoute(this, 'GetHttpRoute', {
+      httpApi: httpApi,
+      integration: apiIntegration,
+      routeKey: HttpRouteKey.with('/{proxy+}', HttpMethod.GET),
+    });
+
```
williamputraintan, line 150: @mmalenic Wanted to confirm from the [FM docs](https://github.com/umccr/orcabus/blob/main/lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md), FM only used GET and PATCH method?

williamputraintan:
Apparently, APIGW JWT Authorizer works best where we define each method instead of `ANY`. It seems that defining this way allows the `OPTIONS` without an `authorizer` header included, whereas ANY require all methods to have an `auth` header.

skel/django-api/deploy/stack.ts
```diff
@@ -116,20 +112,15 @@ export class ProjectNameStack extends Stack {  // FIXME change construct name
...
-      apiName: 'SequenceRunManager',
-      customDomainNamePrefix: 'sequence',
-      ...props,
-    });
+    const srmApi = new ApiGatewayConstruct(this, 'ApiGateway', props.apiGatewayCognitoProps);
     const httpApi = srmApi.httpApi;
 
     const apiIntegration = new HttpLambdaIntegration('ApiIntegration', apiFn);
 
-    new HttpRoute(this, 'HttpRoute', {
```
williamputraintan, line 129: @victorskl @reisingerf On the SRM and WFM only the `GET` method currently used?

victorskl: Read only `GET` for now.

victorskl (approved):

reisingerf (approved):

lib/workload/stateless/stacks/filemanager/deploy/stack.ts
mmalenic: There's a `POST` route as well on `/api/v1/ingest` which manually ingests data from an SQS queue. It isn't being used right now, but could you also please add it here?

mmalenic (approved):

lib/workload/stateless/stacks/filemanager/deploy/stack.ts
williamputraintan: Resolved - ccf38b2de940bba17076993616dade51fa375d78

PR #534, opened 2024-09-05
refactor(filemanager): compile database and migrations

### LINKED ISSUE #530: filemanager: remove database requirement to compile code
We should consider removing the requirement on a postgres database to compile code. While this was initially a good feature in early development, I think it's not that useful anymore and tends to get in the way as most tools just don't expect code compilation to require database access. See https://github.com/umccr/orcabus/pull/429 and https://github.com/umccr/orcabus/pull/529. 

This would involve turning any sqlx `query!` calls into function calls using `query()`, as these don't require database access. It would also mean removing the `filemanager-build` sea-orm entity generation step and instead adding the generated entities into the code base.

I think this change would be beneficial, as it would reduce the hassle of compiling the filemanager codebase and also probably improve compile times.

mmalenic:
Closes #530

### Changes
* A running postgres database is no longer required to compile filemanager (reducing some compile times and hopefully repetitive compilations too).
    * sqlx queries no longer use macros and instead use runtime functions.
    * sea-orm entities are still generated by `filemanager-build`, however as a separate step to `build.rs`. The entities are checked into the codebase, and are assumed to be up to date when running deployment related steps. When developing in the filemanager directory, a database connection is still required to refresh the entities.
* Filemanager migrations are squashed.


mmalenic:
This PR should close https://github.com/umccr/orcabus/pull/532

victorskl (approved):
LGTM

PR #540, opened 2024-09-12
Change(WorkflowRunManager): add more query and search terms

raylrui:
1. add status query params for filter by the latest status
2. add libraryId and orcabusId in the search terms for group workflow runs

example: 
`http://localhost:8000/api/v1/workflowrun/?search=lib.01J5S9CFX5P69S4KZRQGDFKV1N&status=GENERATING_OUTPUTS`
`http://localhost:8000/api/v1/workflowrun/?search=L2400255&status=READY`

### DIFF
--- a/lib/workload/stateless/stacks/workflow-manager/workflow_manager/viewsets/workflow_run.py
+++ b/lib/workload/stateless/stacks/workflow-manager/workflow_manager/viewsets/workflow_run.py
@@ -52,13 +52,9 @@ class WorkflowRunViewSet(ReadOnlyModelViewSet):
             'end_time',
             'is_ongoing',
             'status',
+            'search'
         ])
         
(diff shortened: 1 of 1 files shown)


reisingerf (approved):
Not sure if we'll run into clashes / limits with "L" for libraries, but happy to give it a go.

raylrui:
yes, I am also think of that, what I thought is we can combine all search results in different scope and return them together. 
I will update that later after some testing...

reisingerf:
All good!

raylrui:
default "fuzzy" search" will be removed, and new search terms will constraint to (worfkflow run name, comment, library_id, orcabus_id, workflow name), which will be defined by us

In UI, there will be a tooltips to show available search items
image

raylrui:
No model or migrate added. Merge now ......

PR #542, opened 2024-09-13
Fix (MM): Library synchronisation issue

williamputraintan:
Since sync usually specifies per year level. It will now only allow deletion records in the library model. The deletion will also have additional criteria where it only compares libraries based on the library prefix year. 

This PR is changing the behaviour described in this part of the readme:

```
2. The tracking sheet is the single source of truth for the current year. Any deletion or update to existing records
   will be applied based on their internal IDs (`library_id`, `specimen_id`, and `subject_id`). For the library
   model, the deletion will only occur based on the current year's prefix. For example, syncing the 2024 tracking
   sheet will only query libraries with `library_id` starting with `L24` to determine whether to delete it.
```


victorskl (approved):
Thanks for quick fix, Will. LGTM. Let give this a shot.!

reisingerf (approved):
Sound ok to me

PR #552, opened 2024-09-15
Add lane to bclconvert samplesheet if it doesn't exist

alexiswl:
If there is no lane attribute in the samplesheet bclconvert_data section, collect the runinfo xml and duplicate every row in the samplesheet for every lane in the instrument run


victorskl (approved):
LGTM. Minor unit test case to attend.

alexiswl:
Ah looks like more-itertools isn't a builtin, will add to the requirements tomorrow

PR #553, opened 2024-09-16
Set cttso v2 nf outputs

### LINKED ISSUE #547: cttso get outputs sfn should override status of successful workflow if errors have occurred
ICAv2 exports a success workflow status, despite the one and only sample failing.  

In this case, it's pretty harmless since it's just the NTC and it failed because an evidence bam wasn't generated (due to the lack of reads).  

When there's a failure, we can tell because: 

* The 'errors' folder exists adjacent to 'Results', 'Logs_Intermediates' and 'TSO500_Nextflow_Logs'
* The 'MetricsOutput.tsv' contains the 'COMPLETED_ALL_STEPS' attibute under 'Analysis Status' as 'FALSE'

alexiswl:
Resolves #547


victorskl (approved):
LGTM

victorskl:
I will yank the `0.2.1` release tag to include this too, Alexis. FYI

PR #555, opened 2024-09-16
Initialise library set to prevent sfn failing

### LINKED ISSUE #544: wgts qc kwik-glue sfn fails when no wgts qc samples are on the run
We get a failure because no dynamodb attribute library_id_set has been set for this instrument run id.  

Can either place a 'choice' inside the trigger to determine if the item is present or initialise an empty set first

alexiswl:
sfn event-to-ready will fail when no library has been initialised for this sample type on this run

Resolves #544


victorskl (approved):
LGTM.

Same, I will yank this patch into `0.2.1`.

PR #565, opened 2024-09-20
fix(filemanager): swap docker hub images for AWS public ECR

### LINKED ISSUE #564: filemanager: avoid pulling images from docker hub
The AWS public ECR should be preferred for docker images, as docker hub gets throttled on CodeBuild which causes builds to fail. The `cargo-chef` in the filemanager `Dockerfile` image can be replaced with the base ECR Rust image, and `cargo-chef` can be [installed](https://github.com/LukeMathWalker/cargo-chef?tab=readme-ov-file#without-the-pre-built-image) using `cargo`.

mmalenic:
Closes #564


victorskl (approved):
Rocket!

PR #568, opened 2024-09-22
refactor(filemanager): match AWS behaviour for delete markers

### LINKED ISSUE #563: filemanager: reconsider delete marker record deleted/created status
Currently file manager tracks delete markers created as `Created` event records and delete markers deleted as `Deleted` records. However, AWS sends `Deleted` before both creating and deleting a delete marker. The file manager should probably match this behaviour in order to avoid confusion, and to allow `current_state` queries to correctly determine which object is current, without having to use the `is_delete_marker` column.

mmalenic:
Closes #563

### Changes
* Delete markers are ingested as `Deleted` events, rather than `Created` events, which matches what AWS does.
* Fix `filemanager-build` issue which was attempting to parse non Rust files.
* Update delete marker tests.

The justification behind this is that filemanager should try to match what AWS does so that there are no surprises. It is also helpful in ensuring that `current_state` queries don't return `delete_marker` objects which don't represent the current state.


victorskl:
.. will review in tick, Marko.

victorskl (approved):
LGTM

PR #585, opened 2024-10-03
feat(filemanager): ingest_id tagging and object move tracking

### LINKED ISSUE #584: feat(filemanager): track how an object moves
The filemanager should be able to track how an object moves between bucket/key locations on S3. Since S3 only uses `Created` and `Deleted` events, one approach is to tag S3 objects with an id that the filemanager can use to track it, as tags get copied by default when an object is copied. The filemanager should also be able to copy attributes across records when an object is moved/copied.

mmalenic:
Closes #584

### Mechanism
* Introduces a mechanism by which the filemanager can track how an object moves by using S3 tags.
    * For each object, a tag is added to S3 to track the `ingest_id`.
    * The ingest id gets copied to the database as well.
    * When an object is moved with tags, the `ingest_id` is reused, which allows creating the sequence of records representing the moved object.
    * Attributes are also copied as the object moves.

### Implementation
* Added `GetObjectTagging` and `PutObjectTagging` capabilities to filemanager.
* Added `ingest_id` column to database which matches the `ingest_id` on the S3 object tags.
* Test cases for move logic.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/docs/MOVED_OBJECTS.md
+++ b/lib/workload/stateless/stacks/filemanager/docs/MOVED_OBJECTS.md
@@ -5,9 +5,9 @@ S3 events alone, it's not possible to tell whether an object that has been delet
 the same object that has been moved, or two different objects. This is because S3 only tracks `Created` or `Deleted`
 events.
 
-To track moved objects, the filemanager stores additional information in S3 tags, that gets copied when the object
+To track moved objects, filemanager stores additional information in S3 tags. The tag gets updated when the object
 is moved. This allows the filemanager to track how objects move and also allows it to copy attributes when an object
-is moved/copied.
+is moved/copied. This process is described [below](#tagging-process).
 
 ## Tagging process
 
@@ -26,14 +26,18 @@ API provides a way to query the database for records with a given `ingest_id`.
 
 ## Design considerations
 
-Object tags on S3 are limited to 10 tags per object, and each tag can only store 258 unicode characters. The filemanager
-avoids storing a large amount of data by using a UUID as the value of the tag, which is linked to object records that
-store attributes and data. 
+Object tags on S3 are [limited][s3-tagging] to 10 tags per object, and each tag can only store 258 unicode characters.
+The filemanager avoids storing a large amount of data by using a UUID as the value of the tag, which is linked to object
+records that store attributes and data. 
 
 The object tagging process cannot be atomic, so there is a chance for concurrency errors to occur. Tagging can also
 fail due to database or network errors. The filemanager only tracks `ingest_id`s if it knows that a tag has been
 successfully created on S3, and successfully stored in the database. If tagging fails, or it's not enabled, then the `ingest_id`
 column will be null.
+
+The act of tagging the object allows it to be tracked - ideally this is done as soon as possible. Currently, this happens
+at ingestion, however this could have performance implications. Alternative approaches should consider asynchronous tagging.
+For example, [`s3:ObjectTagging:*`][s3-tagging-event] events could be used for this purpose.
 
 The object tagging mechanism also doesn't differentiate between moved objects and copied objects with the same tags.
 If an object is copied with tags, the `ingest_id` will also be copied and the above logic will apply.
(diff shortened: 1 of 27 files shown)


reisingerf:
Did you test that with any of the BYOB buckets? 
I suppose this will require adjustment to the access permissions / bucket policies. Do you want to PR that against the infrastructure repo?
(i.e. [this](https://github.com/umccr/infrastructure/blob/23e7526efa9b71273be9477325bfa25904a56b35/terraform/stacks/unimelb/data_archive/byob_ica_v2.tf#L137) statement for each BYOB)

mmalenic:
> Did you test that with any of the BYOB buckets? 
> I suppose this will require adjustment to the access permissions / bucket policies. Do you want to PR that against the infrastructure repo?

Good point, I haven't tested on BYOB - I'll PR that on the infra repo.

lib/workload/stateless/stacks/filemanager/docs/MOVED_OBJECTS.md
```diff
@@ -0,0 +1,47 @@
...
+
+The object tagging process cannot be atomic, so there is a chance for concurrency errors to occur. Tagging can also
+fail due to database or network errors. The filemanager only tracks `ingest_id`s if it knows that a tag has been
+successfully created on S3, and successfully stored in the database. If tagging fails, or it's not enabled, then the `ingest_id`
+column will be null.
+
+The object tagging mechanism also doesn't differentiate between moved objects and copied objects with the same tags.
+If an object is copied with tags, the `ingest_id` will also be copied and the above logic will apply.
+
+## Alternative designs
```
reisingerf, line 41: NIce doc!

I'd probably add another small note on the checksum approach: it can't be used if the checksums are not expected to be the same, e.g. with compression, which is a big use case for us.

lib/workload/stateless/stacks/filemanager/docs/MOVED_OBJECTS.md
```diff
@@ -0,0 +1,47 @@
+# Tracking moved objects
+
+The filemanager tracks records from S3 events, which do not describe how objects move from one location to another. Using
+S3 events alone, it's not possible to tell whether an object that has been deleted in one place and created in another is
+the same object that has been moved, or two different objects. This is because S3 only tracks `Created` or `Deleted`
+events.
+
+To track moved objects, the filemanager stores additional information in S3 tags, that gets copied when the object
```
brainstorm, line 8: doc nit:

"stores additional information in S3 tags. The tag field X gets updated when the object is moved."

brainstorm: and/or "see below for tag key-value details"

lib/workload/stateless/stacks/filemanager/docs/MOVED_OBJECTS.md
```diff
@@ -0,0 +1,47 @@
...
+
+* When an object record is ingested, the filemanager queries `Created` events for tags. 
+* If the tags can be retrieved, the filemanager looks for a key called `ingest_id`. The key name can be 
+  configured in the environment variable `FILEMANAGER_INGESTER_TAG_NAME`.
+* The tag is parsed as a UUID, and stored in the `ingest_id` column of `s3_object` for that record.
+* If the tag does not exist, then a new UUID is generated, and the object is tagged on S3 by calling `PutObjectTagging`. 
+  The new tag is also stored in the `ingest_id` column.
+* The database is also queried for any records with the same `ingest_id` so that attributes can be copied to the new record.
+
+This logic is enabled by default, but it can be switched off by setting `FILEMANAGER_INGESTER_TRACK_MOVES`. The filemanager
```
reisingerf, line 24: Tagging is currently part of the ingestion process, right? 
So there's a possibility that this may slow down the ingestion and may become an issue under heavy load?
Not for now, but if that should become the case, we could think of an async tagging strategy. 
Given the option to disable tagging (or tagging failing/missing for other reasons), it would be great to think of an async tagging option.

reisingerf (approved):
Sounds good to me

brainstorm (approved):

lib/workload/stateless/stacks/filemanager/docs/MOVED_OBJECTS.md
mmalenic: Yeah I agree with this, it could potentially slow things down as it's done on ingestion. It's a bit tricky because the act of tagging the object conveys the information of the move - ideally this would be done as soon as possible (i.e. on ingestion). Anything async would extend the window that the object isn't tagged, meaning that the move can't be tracked. In practice this probably wouldn't make a different if the object isn't moved as soon as it's created.

There are `s3:ObjectTagging:*` events which might be good for this.

victorskl: This is always challenging and tradeoff. Let's give it a shot!

victorskl (approved):

lib/workload/stateless/stacks/filemanager/docs/MOVED_OBJECTS.md
reisingerf: Yes, tricky but that's a general "issue" with event bases systems: there's an inevitable delay/asynchronicity.

And I am not saying we should implement that now. An open ticket or comment in the code to keep track of it is perfectly fine.

To compensate for potential concurrency issues, the mentioned support of checksums, name matches, etc could be used... at least to some extend. All future considerations... all good for now!

PR #587, opened 2024-10-03
fix(filemanager): version_id for S3 API calls

### LINKED ISSUE #586: fix(filemanager): use version_id when calling S3 APIs
The filemanager should use `version_id` when calling S3 APIs like `HeadObject`, `GetObjectTagging`, and `PutObjectTagging`. Even though S3 is called on ingestion where the latest object for a `key` and `bucket` is probably always the right object, using the `version_id` is more correct, and it is available at that point in the code.

mmalenic:
Closes #586

Depends on #585

### Changes
* Use version_ids for `HeadObject`, `GetObjectTagging`, `PutObjectTagging` and presigned `GetObject` requests.
    * If versioning is not enabled, then the default `"null"` string version is used, which lines up with how AWS treats buckets without versioning enabled.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/deploy/constructs/functions/api.ts
+++ b/lib/workload/stateless/stacks/filemanager/deploy/constructs/functions/api.ts
@@ -25,6 +25,6 @@ export class ApiFunction extends fn.Function {
       ...props,
     });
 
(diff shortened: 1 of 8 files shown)


brainstorm (approved):

victorskl (approved):
LGTM

PR #588, opened 2024-10-04
Fix error when checking source uris have been transferred

alexiswl:
Quick fix for bssh fastq copy error (will also need to apply patch to the cttsov2 pipeline manager since this also uses the same copy technique)


victorskl (approved):
Thanks for quick fix, Alexis. I will roll this out as hotfix patch release 0.2.3 to prod.

victorskl:
Merging...

PR #589, opened 2024-10-04
Friday changes:

alexiswl:
## PierianDx

* Added pieriandx pipeline manager (monitors runs and sends events once the reports are generated)
* Added nails for stacky to connect cttsov2 to pieriandx pipeline

## Workflow Type Renames

* wgtsQc to wgts-qc
* tumor_normal to tumor-normal

## ICAv2 Resilience

* Allow a ‘rerun’ of a workflow execution if the analysis id doesn’t exist (but the portal id is in the database).

* Reduce concurrency in cttsov2 copy jobs to 5

* Reduce concurrency in bsshfastqcopy jobs to 5

## Metadata

* Refactored metadata tools to match new metadata manager structure in dev
* Made metadata mapper lambda construct that allows user to easily collect an orcabus id from an internal id (library id / subject id / project id etc) and vice versa.

## Showers

Don’t produce subject based showers.

## Stacky

Refactored stacky services to not register subjects / instruments, libraries only, use a scan to find other libraries that have the same subject / on same instrument instead. Reduces number of components within the stack.


victorskl:
Massive changes.! Going through today..

reisingerf:
Uh... way too much for me to go through in detail, so will focus on a few points of interest...

victorskl:
Hey I have 3 coffee ☕ today. All double shots! So why not.

lib/workload/stateless/stacks/pieriandx-pipeline-manager/Readme.md
```diff
@@ -0,0 +1,247 @@
+# PierianDx Pipeline Manager
+
+Giant, serverless pipeline attached to a serverless database
+
+## Using the AWS Launch Step function
+
+The AWS Launch step function takes in the following inputs
+
+* dag: Object
```
reisingerf, line 9: eh, what's a `dag`?

lib/workload/stateless/stacks/pieriandx-pipeline-manager/Readme.md
```diff
@@ -0,0 +1,247 @@
+# PierianDx Pipeline Manager
+
+Giant, serverless pipeline attached to a serverless database
```
reisingerf, line 3: This does not create WRSC events, right? 
Would it make sense to add those, so the WorkflowManager can track them?

(I might have missed it)

reisingerf (approved):
Comments, but no blocker from my side...

lib/workload/stateless/stacks/stacky-mcstackface/glue-constructs/jb-weld/part_3/fastq-list-row-event-shower-complete-to-cttsov2-ready/step_functions_templates/fastq_list_row_shower_complete_event_to_cttsov2_draft_events_sfn_template.asl.json
```diff
@@ -11,25 +11,51 @@
     },
     "Get Libraries from Instrument Run": {
       "Type": "Task",
-      "Resource": "arn:aws:states:::dynamodb:getItem",
+      "Resource": "arn:aws:states:::aws-sdk:dynamodb:scan",
```
victorskl, line 14: Just a note.

This is ok. Perhaps, create GSI/LSI indexes on those none PK/SK lookup attributes where appropriate.
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/SecondaryIndexes.html

If, however, the use case started to explode such that the need of `select * from table` use case all the time or, some dynamic with filtering and predicates condition to a table then it is ok to tap into Postgres.

Anyhow. We shall observe this part as-is for now.

victorskl (approved):
LGTM. Massive effort!

config/constants.ts
```diff
@@ -291,7 +291,7 @@ External resources required by the wgtsqc Stack
 // Deployed under dev/stg/prod
 export const wgtsQcIcav2PipelineIdSSMParameterPath = '/icav2/umccr-prod/wgts_qc_4.2.4_pipeline_id'; // 03689516-b7f8-4dca-bba9-8405b85fae45
 
-export const wgtsQcIcav2PipelineWorkflowType = 'wgtsQc';
+export const wgtsQcIcav2PipelineWorkflowType = 'wgts-qc';
 export const wgtsQcIcav2PipelineWorkflowTypeVersion = '4.2.4';
```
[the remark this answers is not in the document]
alexiswl: Ah let me fix bsshFastqCopy too, bsshFastqCopy is not used in any output paths so got less attention but I think it would be good to fix for workflow run name consistency

lib/workload/stateless/stacks/pieriandx-pipeline-manager/Readme.md
alexiswl: The 'nails' glue creates workflow run state change 'ready' events.   

The monitor-pieriandx-runs step functions will generate a workflow run state change event once pieriandx has completed

config/constants.ts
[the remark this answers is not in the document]
reisingerf: btw: are we OK with `cttsov2`?

First, we are adding a version where the workflow definition has already a `version` field. That feels redundant and confusing and may limit our future uses of that workflow (e.g. what if there's a new version 3.0.1?)
Second, we've seen now multiple other places where the analysis and the sample type are separated, e.g. tso-ctdna. I wonder if we are limiting the possible use cases combining those into one term, e.g. could there be a tso-ctrna, a tso600-ctdna or others? 

I just want to take this opportunity to rethink our workflow naming based on the experience we have now, not simply following the convenience of what we used so far. 
I am not the best person to judge this though, so I'll follow the consensus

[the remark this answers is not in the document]
pdiakumis: Maybe a discussion on Slack where we can reach a consensus.

lib/workload/stateless/stacks/stacky-mcstackface/glue-constructs/jb-weld/part_3/fastq-list-row-event-shower-complete-to-cttsov2-ready/step_functions_templates/fastq_list_row_shower_complete_event_to_cttsov2_draft_events_sfn_template.asl.json
alexiswl: This is runs one per instrument run, I don't think this is a burden on our compute resources

PR #596, opened 2024-10-08
feat(filemanager): accept multiple keys in API

### LINKED ISSUE #595: filemanager: allow API to accept multiple filter parameters with the same key
The filemanager API should allow accepting multiple filter parameters of the same key, where each key represents an `or` statement in the SQL query.

For example:
```sh
/api/v1/s3?bucket[]=bucket1&bucket[]=bucket2
```

Should resolve to an SQL query that fetches records with either `bucket1` or `bucket2` as the bucket:
```sql
select * from s3_object where bucket = 'bucket1' or bucket = 'bucket2';
```

The extra `[]` in the query parameters represents multiple keys and is required by `serde_qs`.

mmalenic:
Closes #595 

### Changes
* The filemanager API can now accept multiple keys in the API, where each additional key represents an `or` statement in the SQL query.
    * Additional keys need to use the `[]` syntax to specify a list. E.g. `/api/v1/s3?key[]=key1&key[]=key2`.
* Multiple keys are also supported on attributes.
    * The logic involving attributes has been fixed so that JSON arrays represent an `or` condition.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md
+++ b/lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md
@@ -149,9 +149,8 @@ or `bucket2`:
 curl -H "Authorization: Bearer $TOKEN" "https://file.dev.umccr.org/api/v1/s3?bucket[]=bucket1&bucket[]=bucket2" | jq
 ```
 
-Note that the extra `[]` is required in the query parameters to specify multiple keys with the same name. Multiple keys
-are also supported on attributes. For example, the following finds records where the `portalRunId` is either `20240521aecb782`
-or `20240521aecb783`:
+Multiple keys are also supported on attributes. For example, the following finds records where the `portalRunId` is
+either `20240521aecb782` or `20240521aecb783`:
  
 ```sh
 curl --get -H "Authorization: Bearer $TOKEN" \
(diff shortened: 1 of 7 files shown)


lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md
```diff
@@ -139,6 +139,27 @@ curl --get -H "Authorization: Bearer $TOKEN" --data-urlencode "attributes[portal
...
+
+The API supports querying using multiple keys with the same name. This represents an `or` condition in the SQL query, where
+records are fetched if any of the keys match. For example, the following finds records where the bucket is either `bucket1`
+or `bucket2`:
+
+```sh
+curl -H "Authorization: Bearer $TOKEN" "https://file.dev.umccr.org/api/v1/s3?bucket[]=bucket1&bucket[]=bucket2" | jq
+```
+
+Note that the extra `[]` is required in the query parameters to specify multiple keys with the same name. Multiple keys
```
williamputraintan, line 152: What is the reason behind it where it needs to have the `[]` for a multiple key? Will this be mutually exclusive if there is a params that doesn't have `[]`? Say `...s3?bucket=bucket1&bucket[]=bucket2`

mmalenic: It's required in the library that parses the query string: https://github.com/samscott89/serde_qs. I think the choice by the author has to do with matching the style of ruby on rails. 

Yes, this is mutually exclusive, you can't combine keys with `[]` and without it.

For example, this works:
1. `/api/v1/s3?key=...`
2. `/api/v1/s3?key[]=...`
3. `/api/v1/s3?key[]=...&key[]=...`

But this results in an error:
1. `/api/v1/s3?key=...&key[]=...`

victorskl (approved):

williamputraintan (approved):
LGTM! Thanks for the quick feature request update.

PR #598, opened 2024-10-09
Fix: Drop incomplete record before metadata processing

williamputraintan:
Fixes where it loads records where their internal Ids (e.g. `individualId`) are still null (as it is incomplete entry) by dropping the record entirely.


victorskl:
Related https://umccr.slack.com/archives/C03ABJTSN7J/p1727745957473829

victorskl (approved):

reisingerf (approved):

PR #601, opened 2024-10-11
feat(filemanager): docker caching and current state default

### LINKED ISSUE #600: filemanager: support sccache inside Dockerfile
The filemanager should use [sccache](https://github.com/mozilla/sccache) inside the Dockerfile with a cache mount so that builds across different contexts (e.g. local Makefile, CDK, testcontainers) can all utilize the same cached compilation targets.

mmalenic:
Closes #600

### Changes
* Adds sccache to the filemanager `Dockerfile`, which uses a buildx cache mount to utilize a global cache across different build contexts (like CDK synth, or the fmannotator testcontainers code).
    * This should hopefully improve repeated builds of the filemanager, because each docker builld will use the same cache mount.
* Makes `currentState` true by default in the filemanager API.
* Fixes an issue with the `/s3/attributes` route which wasn't taking into account known parameters like `currentState` or `page`.
* Use docker compose in fmannotator tests because it properly supports rebuilding images and not deleting them when `docker compose down` is called.


brainstorm (approved):
Kudos, TIL about `cargo binstall` 👍🏻

victorskl (approved):

PR #602, opened 2024-10-14
Add fastqc / sequali stats to fastq list row events for first 1 million reads

alexiswl:
Write information to tables to prevent overfill of data in step functions object


victorskl (approved):

alexiswl:
Ah @victorskl I don't think this can be cherry picked for O2, the underlying dynamodb uses the #589 syntax of the library orcabus id being the primary key for this particular database.  So as you suggested, run 333 will need to be run once we're on O3

victorskl:
Yup. No blocker on Run 333 at ICA v2 side. 
Happy to put into backlog and I will follow up once O3 concluded.

PR #603, opened 2024-10-14
cttsov2 job checker requires permissions to list executions

alexiswl:
Error created when cdk-nag forced us to use our own policy statement instead of 'grantRead', missed this permission


victorskl:
Related #589

PR #606, opened 2024-10-16
Feat (MM): History API for Models

### LINKED ISSUE #402: Mechanism to query Metadata History
The metadata-manager will track daily changes if any. This is done with [django-simple-history](https://django-simple-history.readthedocs.io/en/latest/) framework, perhaps we could elevate the query mechanism from this framework to be able to view the changes.

williamputraintan:
- sync mock data between `orcabus_id`/`library_id` between WFM and MM (useful for UI purposes)
- Added history API for each model

Resolves #402


lib/workload/stateless/stacks/metadata-manager/app/management/commands/insert_mock_data.py
```diff
@@ -1,22 +1,166 @@
...
+        {
+            "orcabus_id": "06J5M2JFE1JPYV62RYQEG99CP6",
+            "phenotype": "tumor",
+            "library_id": "L000006",
+            "assay": "ctTSOv2",
+            "type": "ctDNA",
+            "subject": "SBJ00003",
+            "workflow": "research"
+        },
+    ]
```
victorskl, line 93: Just dropping a note. No action need for the comment.

We should reorganise this as centralised seed data (in JSON or YAML format) somewhere in `docs` or `shared` directory down the track. It will get complex; when we include scenario and more services in the role play -- Case Manager, etc.

For current O3 iteration, this is fine.

victorskl (approved):

reisingerf (approved):

Issue #605, opened 2024-10-14
filemanager: lifecycle transitioning objects


mmalenic:
The filemanager should support a mechanism to transition objects to an expired state or a different storage class. This involves accepting a transition request via the API or the EventBus for a set of objects, and then performing the transition after the specified time period.

I think this can be accomplished with object tagging, and S3 lifecycle rules. For example, the filemanager could tag objects with predefined values matching lifecycle rules. This should only require one tag per transition. Moving should also not be an issue, because the filemanager can remove the lifecycle tag when an object is moved, as it's tracking moves via the `filemanager_id` tag.

A limitation of this approach is that it requires additional tags for each transition, and there is a maximum of 10 tags per objects. It doesn't look like there is a way to pack tag values into a single key to support multiple transitions, because the transition rules expect exact values for tags.

As an alternative, the filemanager could control the lifecycle transitions using it's own database state and `DeleteObject`. This could be more complicated but would remove the tag requirement, and allow supporting transitions using more complex logic like wildcards.

mmalenic:
It is possible to specify multiple transitions in a single lifecycle rule, so it should be possible to define rules that transition more than once based on a single tag. For example, a tag with the value `standard_ia_30&glacier_90` could match a rule with a transition for standard_ia after 30 days and glacier after 90 days.

mmalenic:
Transferred to https://github.com/OrcaBus/service-filemanager/issues/28

PR #607, opened 2024-10-17
feat(filemanager): add `FileStateChange` event schema

mmalenic:
Related to #605, addresses point 1 of https://github.com/umccr/orcabus/issues/605#issuecomment-2418388645.

### Changes
* Adds the `FileStateChange` schema which will support tagging objects to transition them to another storage class, or delete them.
    * I don't think I've missed anything, as both expiring and transitioning objects is supported.
    * I haven't added any support for `NoncurrentVersionTransition`, `NoncurrentVersionExpiration`, `AbortIncompleteMultipartUpload`, or `ExpiredObjectDeleteMarker`, because I feel like these would be covered by bucket-wide rules that apply to all objects? It doesn't seem like it would be useful to have these apply differently based on the object prefix. I'm happy to add it in if we think it's useful though?
* The implementation uses a build command to generate the schema from Rust code. This is more convenient because the same code can be used to parse incoming events.


brainstorm (approved):

reisingerf:
Hm, not sure I understood the intention correctly... 

The schema defines a **FILE** state change event, but it does not really link it to any particular file. It rather feels like a policy or lifecycle transition rule change. 

It also does not seem to have provisioning for other "state" changes, like "archived" or "deleted" or "moved" .... 

The "source" in the examples is `orcabus.filemanager` indicating an announcement from the FM?

Also, not sure all the S3 lifecycle details are needed/useful. We can't create rules for each and every file/event...

mmalenic:
Yes, I think this probably isn't a `StateChange` kind of event, it more a transition event.

> It also does not seem to have provisioning for other "state" changes, like "archived" or "deleted" or "moved" ....

On this note, one thing the filemanager could do that isn't possible using s3 lifecycles is to move objects between locations. E.g. a workflow run could say to move the objects to an archive bucket after n-days. Not sure if something like this would be useful.

victorskl:
@reisingerf @mmalenic Let us quick catch up call next week on this, pls. I'll follow up ping you on next Monday.

mmalenic:
Putting this feature on hold for now, as we need to figure out if using events/tags is the best approach for this.

PR #613, opened 2024-10-23
wfr/ feat-comments-resolved-status-mutation-module

raylrui:
1. wfr comment module, with CRUD func
2. resolved state module, with create and update (comment) func
3. pagination class to None for state and comment, as we need all state and comment at once
4. modify extend_schema, as it may mark required field for list call. New custom mixin class can be add to remove all required field for list call
```
 @extend_schema(parameters=[
        xxxxListParamSerializer
    ])
```
5. couple with frontend PR https://github.com/umccr/orca-ui/pull/51
link issue: https://github.com/umccr/orca-ui/issues/55


williamputraintan:
> 4. remove extend_schema, as it changes request params
> 	```
>  @extend_schema(parameters=[
>        AnalysisSerializer
>    ])
>```

This is meant to change the request parameters for the OpenAPI, is this the concern for the deletion? The base manager supports filtering based on its value, so documenting in the OpenAPI should be useful. Maybe some recursive property should be excluded as the filter might not work.

raylrui:
@williamputraintan 
when it added to the list request, it will change the list schema request params like this:
image
but normally we just need this:
image

williamputraintan:
> @williamputraintan when it added to the list request, it will change the list schema request params like this: image but normally we just need this: image
> 
> I am not sure if my open-api crash some way, will check with you

Yes, but `analysisName` is a valid queryParams to filter the records? Perhaps we need to modify it to remove the *required* and some exclusion field where it asks for a non-string object

raylrui:
>Yes, but analysisName is a valid queryParams to filter the records? Perhaps we need to modify it to remove the required and some exclusion field where it asks for a non-string object

but it will make the "analysisName" as required params whenever need to call it

williamputraintan:
> > Yes, but analysisName is a valid queryParams to filter the records? Perhaps we need to modify it to remove the required and some exclusion field where it asks for a non-string object
> 
> but it will make the "analysisName" as required params whenever need to call it

Yes, thats what I meant we need to modify the `required` from the queryParams. We might need to find a way how to do it, instead of completely removing it.

raylrui:
> > > Yes, but analysisName is a valid queryParams to filter the records? Perhaps we need to modify it to remove the required and some exclusion field where it asks for a non-string object
> > 
> > 
> > but it will make the "analysisName" as required params whenever need to call it
> 
> Yes, thats what I meant we need to modify the `required` from the queryParams. We might need to find a way how to do it, instead of completely removing it.

ah, yes, that will be great if 'required' can be removed in swagger ui. But, anyway, it actually don't affect api call, as it can be still callable without these params. just affect the swagger ui, I think.....

raylrui:
> > > Yes, but analysisName is a valid queryParams to filter the records? Perhaps we need to modify it to remove the required and some exclusion field where it asks for a non-string object
> > 
> > 
> > but it will make the "analysisName" as required params whenever need to call it
> 
> Yes, thats what I meant we need to modify the `required` from the queryParams. We might need to find a way how to do it, instead of completely removing it.

it seems not easy to get rid of "required" field as it comes from default Serializer, I will have a check if we can make a another custom Serializer or make some manual params like this:
```
 @extend_schema(
        parameters=[
            OpenApiParameter(name='field1', description='Description of field1', required=False, type=OpenApiTypes.STR),
            OpenApiParameter(name='field2', description='Description of field2', required=False, type=OpenApiTypes.INT),
            # Add more parameters as needed
        ]
    )
```

raylrui:
@williamputraintan 
I build a custom Mixin to mark all field optional when used to list api call, to avoid the required field in orignal model
```
class OptionalFieldsMixin:
    def make_fields_optional(self):
        # Make all fields optional
        for field in self.fields.values():
            field.required = False

        # If the fields are CharField, you might also want to allow them to be blank
        for field_name, field in self.fields.items():
            if isinstance(field, serializers.CharField):
                field.allow_blank = True

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.make_fields_optional()
```

williamputraintan (approved):

raylrui:
thanks, will merge together with ui for testing in dev

reisingerf (approved):

williamputraintan (commented):
Ah, perhaps the CDK needs to add methods other than `GET`

raylrui:
> Ah, perhaps the CDK needs to add methods other than `GET`

sorry, which CDK do you mean?

reisingerf:
I think William means the CDK script that actually deploys the API.
Somewhere around [here](https://github.com/umccr/orcabus/blob/f91d867d623cc06554a09c8c75306a74bedc710b/lib/workload/stateless/stacks/workflow-manager/deploy/stack.ts#L125-L129).

williamputraintan:
> > Ah, perhaps the CDK needs to add methods other than `GET`
> 
> sorry, which CDK do you mean?

The API Gateway HTTP Route for the workflow-manager in `...wfm/deploy/stack.ts`

raylrui:
> > > Ah, perhaps the CDK needs to add methods other than `GET`
> > 
> > 
> > sorry, which CDK do you mean?
> 
> The API Gateway HTTP Route for the workflow-manager in `...wfm/deploy/stack.ts`

yes, new method need to be added to api gateway. Great thanks !

Issue #614, opened 2024-10-24
filemanager: partition record state


mmalenic:
Queries using the current state logic [here](https://github.com/umccr/orcabus/blob/6f2074ff43c5a31e94b1559b0e5b33eec5cad6fd/lib/workload/stateless/stacks/filemanager/filemanager/src/queries/list.rs#L195-L223) are already a bit slow, and will only get slower as the number of records grows. This is because the filemanager never deletes records, even though records transition from current to historical data.

This can be improved by partitioning the records by historical and current state, which should improve the speed of queries on the current state. In  order to do this a new column should be added that can be used to split the data.

Tasks:
- [x] Add new `is_current_state` column and adjust the API to use it.
- [ ] Partition the records on the new column.
    * This will involve some complexity because the table needs to be recreated. Also, in postgres, globally unique constraints can't apply across partitions which will affect the event de-duplication logic.

mmalenic:
Transferred to https://github.com/OrcaBus/service-filemanager/issues/23

PR #615, opened 2024-10-24
feat: filemanager partitions

mmalenic:
Related to #614, finishes first task.

### Changes
* Add `is_current_state` column to database for later partitioning.
    * Migrates existing data to the correct `is_current_state`.
    * API uses this column now instead of performing the calculation every time.
    * Tests are updated to take this into account.
    
I know this adds some overhead to the ingester, but it is performing pretty well in testing (< a few ms per insert), and I think it's worth it because it reduces the performance issues on the API side. If this ever is a problem for the ingester to handle, then it can always be run as a separate asynchronous process.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/database/queries/api/reset_current_state.sql
+++ b/lib/workload/stateless/stacks/filemanager/database/queries/api/reset_current_state.sql
@@ -1,0 +1,41 @@
+-- Resets the `is_current_state` to false for a set of objects based on the `bucket`, `key`, `version_id`
+-- and `sequencer`. This is used to reset the current state so that a new object can have it's `is_current_state`
+-- set to true.
+
+-- Unnest input.
+with input as (
+    select
+        *
+    from unnest(
+        $1::text[],
+        $2::text[],
+        $3::text[],
+        $4::text[]
+    ) as input (
+        bucket,
+        key,
+        version_id,
+        sequencer
+    )
+),
+-- Select objects to update.
+to_update as (
+    select s3_object_id from input cross join lateral (
+        select
+            s3_object_id
+        from s3_object
+        where
+            input.bucket = s3_object.bucket and
+            input.key = s3_object.key and
+            input.version_id = s3_object.version_id and
+            -- Only the objects older than the input.
+            input.sequencer >= s3_object.sequencer and
+            -- Only need to update current objects.
+            s3_object.is_current_state = true
+        for update
+    ) s3_object
+)
+update s3_object
+set is_current_state = false
+from to_update
+where s3_object.s3_object_id = to_update.s3_object_id;
--- a/lib/workload/stateless/stacks/filemanager/database/queries/api/reset_current_state.sql
+++ b/lib/workload/stateless/stacks/filemanager/database/queries/api/reset_current_state.sql
@@ -20,22 +20,35 @@ with input as (
 ),
 -- Select objects to update.
 to_update as (
-    select s3_object_id from input cross join lateral (
+    select * from input cross join lateral (
         select
-            s3_object_id
+            s3_object_id,
+            -- This finds the first value in the set which represents the most up-to-date state.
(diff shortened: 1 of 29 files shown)


lib/workload/stateless/stacks/filemanager/database/queries/api/reset_current_state.sql
```diff
@@ -0,0 +1,59 @@
...
+            input.version_id = s3_object.version_id and
+            -- This is an optimization which prevents querying against all objects in the set.
+            (
+                -- Only need to update current objects
+                s3_object.is_current_state = true or
+                -- Or objects where there is a newer sequencer than the one being processed.
+                -- This is required in case an out-of-order event is encountered. This always
+                -- includes the object being processed as it's required for the above row-logic.
+                s3_object.sequencer >= input.sequencer
+            )
```
mmalenic, line 52: I hope this makes sense! Let me know if I can improve the explanation or if it would be good to reduce some of the complexity. For example, that last nested `or` condition probably isn't necessary for correctness but it should help performance by not querying the whole dataset.

victorskl (approved):

lib/workload/stateless/stacks/filemanager/docs/ARCHITECTURE.md
```diff
@@ -41,6 +41,18 @@ Within the application code, out of order events are removed within the [events]
...
+
+Since the filemanager database keeps growing as records are never deleted, the current state of records is stored on a
+`is_current_state` column. This column indicates which records represent real objects in S3, and which records are
+historical data that represent previously deleted objects.
+
+This value is computed when events are ingested, and automatically kept up to date. This is done at ingestion because
+the performance impact of determining this is too great on every API call. This does incur a performance penalty for
+ingestion. However, it should be minimal as only other current records need to be considered. This is because records
+only need to transition from current to historical (and not the other way around). Records which are current represent
+a smaller subset of all records, meaning that only a smaller part of the dataset needs to be queried when ingesting.
```
reisingerf, line 54: Not sure I completely follow here.... 
Why do other records need to be queried when a new one is ingested?

>from current to historical (and not the other way around)
How would objects restored from the archive bucket be handled?

Also, perhaps for future consideration, if handling the data within one table get's tricky, consider moving records between tables ? E.g. a current_records vs deleted_records table? (I think the Django history functionality works like that)

mmalenic: > Why do other records need to be queried when a new one is ingested?

A record starts as current, and then some logic needs to make it historical.

For example, say `Created` event "A" comes in for a given key/bucket. At that point in time, "A" has `is_current_state` set to true on ingestion. Then, another `Created` event "B" comes in for the same key/bucket, which represents overwriting that object. Now, "B" has `is_current_state` set to true given that it's the new current state. Something needs to flip `is_current_state` to false on "A" as it's no longer current. This requires querying the database, which is what [this](https://github.com/umccr/orcabus/pull/615#discussion_r1815754904) query does on ingestion.

This doesn't have to happen on ingestion, it could happen in another process some time afterwards. However, I think the cost of doing it at ingestion isn't too high so I've added it to make the state of the records as accurate as soon as possible.

Does that make sense (I'll add this in the docs for the next PR)?

> Also, perhaps for future consideration, if handling the data within one table get's tricky, consider moving records between tables ?

Yeah, agreed, that's the idea with task 2 on https://github.com/umccr/orcabus/issues/614. I'll do this by partitioning the table, which more-or-less creates two separate tables, but allows queries to treat it like one table. When this is set up in postgres, it happens automatically on a particular partition key. This is why I'm adding the `is_current_state` here, because this allows the table to be partitioned on that column.

reisingerf: Ah, yes. I forgot we are dealing with object versions.... 
Querying with bucket + key + version should be quick and efficient... all good!

PR #619, opened 2024-10-27
Remove unused stacky tables

alexiswl:
InputMakerTable and WorkflowTable had uses prior to generation of the workflowmanager which now stores runs.  

Tables no longer needed


victorskl (approved):

victorskl:
@alexiswl pls merge when ready

Issue #322, opened 2024-05-24
filemanager: track events other than `Object Created` and `Object Deleted`


mmalenic:
It might be interesting for the filemanager to also track the Multipart Upload events when multipart upload is used. Currently these events are tracked only when the upload is completed, and non of the intermediate parts are tracked. This would involve some additional database columns, and properly parsing [`s3:ObjectCreated:CompleteMultipartUpload`](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-how-to-event-types-and-destinations.html) events.

mmalenic:
Updating this to also include tracking for events like "Object Storage Class Changed" and "Object Tags Added". Currently, filemanager derives all the information it needs from calls to `HeadObject` and `GetObjectTagging`. However to better track state, and for improved efficiency, other event types should be considered.

Tasks:
- [ ] Track `Object Storage Class Changed`/`Object Access Tier Changed` events.
- [ ] Track `Object Tags Added`/`Object Tags Removed` events.
- [ ] Track `Object Restore Initiated`/`Object Restore Completed`/`Object Restore Expired` events.

PR #621, opened 2024-10-29
feat(event-source): ignore cache directories

mmalenic:
Related to https://github.com/umccr/infrastructure/pull/489

### Changes
* Ignore cache directories under `byob-icav2` when processing incoming EventBridge events, and only forward `Created` and `Deleted` events to filemanager.

### DIFF
--- a/config/stacks/shared.ts
+++ b/config/stacks/shared.ts
@@ -94,9 +94,12 @@ const getEventSourceConstructProps = (stage: AppStage): EventSourceProps => {
     rules: [
       {
         bucket: oncoanalyserBucket[stage],
+        eventTypes: ['Object Created', 'Object Deleted'],
       },
       {
         bucket: icav2PipelineCacheBucket[stage],
+        eventTypes: ['Object Created', 'Object Deleted'],
+        key: [{ 'anything-but': { wildcard: 'byob-icav2/*/cache/*' } }],
       },
     ],
   };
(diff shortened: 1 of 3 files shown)


config/stacks/shared.ts
```diff
@@ -94,9 +94,12 @@ const getEventSourceConstructProps = (stage: AppStage): EventSourceProps => {
     rules: [
       {
         bucket: oncoanalyserBucket[stage],
+        eventTypes: ['Object Created', 'Object Deleted'],
```
reisingerf, line 97: Are those the only event types the FileManager is currently monitoring? 
If yes, that's OK, just be aware that this is a while-listing and you'll miss any other events (which is OK for the portal use case, but you'll have to judge if it's also OK for the FileManager).

victorskl (approved):
LGTM.

ditto Flo comment on the part either filtering on eventTypes or otherwise.

config/stacks/shared.ts
mmalenic: Yep, those are the only ones considered right now. It derives all the other information from `HeadObject` and `GetObjectTagging`. In the future, I can add support for "Object Storage Class Changed" and "Object Tags Added" to avoid using an API call.

victorskl: Sound good. Go for it, Marko. Pls merge when you ready.

mmalenic: Adding to this issue: https://github.com/umccr/orcabus/issues/322#issuecomment-2443343299

PR #626, opened 2024-10-30
Fix library lookup with prefixed orcabusId

reisingerf:
As noted in the test:
- for library lookup and library record generation a "sanitised" `orcabusId` is used (e.g. ULID only, prefix stripped off)
- for the event relay the original values from the input event are forwarded as is to the output event

(not a permanent nor elegant solution)


victorskl (approved):
LGTM, Flo!

alexiswl (approved):

victorskl:
Flo, merging this to cover CI build lead time...

PR #638, opened 2024-11-02
Add analysis size option per microservice

### LINKED ISSUE #637: ORA Compression pipeline needs a larger storage size
Add in an analysis storage size option to the sfn icav2 ready event handler. The analysis storage size can then be an input into the lambda launch function

alexiswl:
One day will allow for per-run

Resolves #637


victorskl (approved):
LGTM

PR #643, opened 2024-11-03
Manually transfer files that are smaller than 8 MiB or dont have an etag

alexiswl:
Resolves discussion in https://github.com/umccr-illumina/ica_v2/issues/208


victorskl (approved):
LGTM for the approach and workaround.
Let us give it a shot.

_(I still feel wired about the situation though, tbh)_

reisingerf (approved):

reisingerf:
Yes, let's hope that gets handled on ILMN side and we don't have to deal with this on our end.

Issue #644, opened 2024-11-03
Lambda Python Log Level basicConfig needs special initialisation


alexiswl:
Python Lambda Logger is a little odd

```
logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)

logger.info("Foo")
```

Won't print 'foo' to the cloudwatch logs, 

[This stack overflow answer explains it a little more](https://stackoverflow.com/a/56579088)

From [this stack overflow answer on the same thread](https://stackoverflow.com/a/45624044), we see the comment that since python 3.8 we can just use the `force=True` on basicConfig.  

From https://docs.python.org/3/library/logging.html#logging.basicConfig: 

_This function does nothing if the root logger already has handlers configured, unless the keyword argument force is set to True._

So action is to set force as 'true' to logging.basicConfig.  

We should probably also use `format='%(asctime)s %(message)s'` to provide timestamps in the cloudwatch logs of when logs were printed

PR #645, opened 2024-11-03
Standardise loggers in python lambda functions

alexiswl:
Resolves https://github.com/umccr/orcabus/issues/644

### DIFF
--- a/lib/workload/components/icav2-copy-files/check_or_launch_job_lambda_py/check_or_launch_job_lambda.py
+++ b/lib/workload/components/icav2-copy-files/check_or_launch_job_lambda_py/check_or_launch_job_lambda.py
@@ -50,8 +50,12 @@ if typing.TYPE_CHECKING:
     from mypy_boto3_secretsmanager import SecretsManagerClient
 
 # Set logging
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger(__name__)
+logging.basicConfig(
+    level=logging.INFO,
+    force=True,
+    format='%(asctime)s %(message)s'
+)
+logger = logging.getLogger()
 
 # Globals
 SUCCESS_STATES = [
--- a/lib/workload/components/python-lambda-fastq-list-rows-to-cwl-input/fastq_list_rows_to_cwl_input_py/fastq_list_rows_to_cwl_input.py
+++ b/lib/workload/components/python-lambda-fastq-list-rows-to-cwl-input/fastq_list_rows_to_cwl_input_py/fastq_list_rows_to_cwl_input.py
@@ -13,8 +13,12 @@ import logging
 from typing import Dict, Union
 
 # Set logger
+logging.basicConfig(
+    level=logging.INFO,
+    force=True,
+    format='%(asctime)s %(message)s'
+)
 logger = logging.getLogger()
-logger.setLevel(logging.INFO)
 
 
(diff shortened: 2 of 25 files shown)


victorskl (approved):
LGTM

lib/workload/components/icav2-copy-files/check_or_launch_job_lambda_py/check_or_launch_job_lambda.py
```diff
@@ -50,8 +50,12 @@
     from mypy_boto3_secretsmanager import SecretsManagerClient
 
 # Set logging
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger(__name__)
+logging.basicConfig(
+    level=logging.INFO,
+    force=True,
```
reisingerf, line 55: This will remove any preconfigured handlers... 
Is that what we want? 

The accepted answer claims that the default AWS handler "might also add some metadata to the record if available", which sounds like it could be useful? 

Happy to accept the change. Longer term we should probably switch to the lambda power tools...

reisingerf (approved):

lib/workload/components/icav2-copy-files/check_or_launch_job_lambda_py/check_or_launch_job_lambda.py
alexiswl: Oh those docs are actually very handy!

Hmm, yes you might be right, from this below, the logger may in fact set the requestId as well.  

```
Using structured JSON logs with Python
If you select JSON for your function’s log format, Lambda will send logs output by the Python standard logging library to CloudWatch as structured JSON. Each JSON log object contains at least four key value pairs with the following keys:

"timestamp" - the time the log message was generated

"level" - the log level assigned to the message

"message" - the contents of the log message

"requestId" - the unique request ID for the function invocation
```

We could rearrange from 

```
import logging
logging.basicConfig(
    level=logging.INFO,
    force=True,
    format='%(asctime)s %(message)s'
)
logger = logging.getLogger()
```

To just

```
import logging
logger = logging.getLogger()
logger.setLevel(logging.INFO)
```

And then we still get the same log levels, just lose the formatting a bit

reisingerf: If that works locally... 
I think the issue was that then on the local env there may not be a log handler at all, hence the fallback to the `basisConfig` if no handler was found.

alexiswl: > I think the issue was that then on the local env there may not be a log handler at all,

Yes, the answer that looks for handlers didn't seem very fun. 

Something worth considering is this too - https://docs.powertools.aws.dev/lambda/python/latest/core/logger/

reisingerf: Yes, the power tools logging support. That's what I meant earlier.
(but there are also other reasons to use the power tools)

victorskl:
Resolve this PR soon too, @alexiswl... 
Opt to upgrade to power tools when appropriate, etc.

alexiswl:
On the backlog, closing PR for now

PR #599, opened 2024-10-10
Feature/analysis journey

reisingerf:
This includes some breaking changes, like the change from internal auto-id fields to OrcaBus IDs as PKs. 

There are also updates to the API with removed/added endpoints and updated/extended response schemas.

PR #646, opened 2024-11-03
Only save payload if present

reisingerf:


victorskl (approved):
Too easy!

victorskl:
Related #599

Issue #653, opened 2024-11-05
PierianDx runs fail as CodePipeline fails to download GitLfs data into lambda layer


alexiswl:

alexiswl:
@williamputraintan we have a workaround in the PR, but wonder if there's a way to get this working within CodePipeline as well

alexiswl:
Not a huge priority though

PR #654, opened 2024-11-05
Pull the snomed databases from github if they're not on the local system

alexiswl:
Workaround until codepipeline supports Git LFS

Workaround for #653


victorskl (approved):
LGTM

PR #659, opened 2024-11-06
PierianDx Production Fixes

alexiswl:
* Prod PierianDx bucket requires KMS encryption
* Can't use LATEST version for invocation of 'collectPierianDxAccessToken'
	* Use '*' on function name followed by NagSuppression
* Fix SSM parameter determining the path of the PierianDx bucket
* Differences between requesting_physicians_x_name and requetsing_physician_x_name was causing default name to be used (Sean)
* Wrong JSON path used in SFN to collect the panel version
* Report ID stored in database as an integer not as a string
* Fixed GET for linked libraries which had an incorrect JSON path when pushing a workflow run state change event
* Fixed analysis path generation for data outputs

P.S have already pushed this to OrcaBus Prod manually


victorskl (approved):
LGTM

PR #617, opened 2024-10-25
Feat: Add Authoriser Based on User Pool Group

williamputraintan:
- Add AWS Verified Permission to OrcaBus to handle authorisation requests. 
- Add lambda authoriser for API-GW to check if the token used is within the admin group of the user pool.

Depends on https://github.com/umccr/infrastructure/pull/487

PR #661, opened 2024-11-07
HotFix(authorization-manager): Sanitize Authorization Header 

williamputraintan:
allow `Bearer ` prefix on the admin authorisation lambda

Related #617


mmalenic (approved):

victorskl (approved):
Sound good.!

williamputraintan:
Also, rename the lambda jwt authoriser so it does not contain the `admin` (policy) naming. The lambda authoriser shouldn't care about the policy (as verified permission should check against it), it should just check the JWT with the given policy Id.

PR #665, opened 2024-11-07
Update dragen somatic pipeline id

alexiswl:
Dragen somatic with germline pipeline didn't have the enable_cnv_somatic flag available in the workflow (and in the past we have just used enable_cnv), since we do not need the cnv on the germline calls we use just enable_cnv_somatic instead.

Updated ssm parameter in dev/stg and prod


victorskl (approved):
LGTM

PR #670, opened 2024-11-08
hotfix(filemanager): add permissions that allow API/inventory functions to access `pipeline-*-cache` buckets

### LINKED ISSUE #669: filemanager: fix presigned URLs for objects under `pipeline-*-cache` buckets
The filemanager API fails to create valid presigned URLs for objects under `pipeline-*-cache` buckets because the API function lacks permissions to access the bucket. This leads to a presigned URL that returns an `Access Denied` error.

mmalenic:
Closes #669 

### Changes
* Gives the API and inventory functions access the `orcabus_file_manager_ingest_access` role. This allows the API function to create valid presigned URLs, and the inventory function to correctly get object metadata before ingesting.


mmalenic:
@williamputraintan This should fix the issue with generating download links in the UI.

victorskl (approved):
LGTM

PR #671, opened 2024-11-08
Wrong typescript boolean logic which meant there was no event rule

alexiswl:
Resolves https://github.com/umccr/orcabus/issues/672


victorskl (approved):
LGTM

PR #686, opened 2024-11-12
Chore(mm): Miscellaneous Update

williamputraintan:
- detailing ER diagram relationship
- Enforce Auth on import metadata from external csv
- Remove libraries returned in the project detail endpoint (as one project could have many libraries)
- Lambda memory based on the[ lambda power tuning](https://github.com/alexcasalboni/aws-lambda-power-tuning)


lib/workload/stateless/stacks/metadata-manager/deploy/construct/lambda-api/index.ts
```diff
@@ -50,35 +51,48 @@ export class LambdaAPIConstruct extends Construct {
...
       'SYNC_CSV_PRESIGNED_URL_LAMBDA_NAME',
       lambdaProps.syncCustomCsvLambda.functionName
     );
+    new HttpRoute(this, 'PostHttpRouteSyncPresignedCsv', {
+      httpApi: apiGW.httpApi,
+      integration: apiIntegration,
+      // Only admin group can use this csv sync endpoint
+      authorizer: apiGW.authStackHttpLambdaAuthorizer,
+      routeKey: HttpRouteKey.with(
+        `/api/${this.API_VERSION}/sync/presigned-csv/{PROXY+}`,
```
williamputraintan, line 93: HTTP API-GW doesn't allow a trailing slash, so it would need `{PROXY+}` here to allow it. We might want to just remove the trailing slash from Django routing completely?

reisingerf: Right!
I'd say for now leave the {PROXY+} in there and we'll re-evaluate a bit down the track?
I have the feeling there are other changes and discussions around the API pending or coming up anyway.

reisingerf (approved):

victorskl (approved):
LGTM

PR #691, opened 2024-11-13
Fix(mm): Broken Comparison on Update

### LINKED ISSUE #690: metadata-manager: Multiple update event for every record
An Update event is sent to the event bus on every record when sync from the tracking sheet

williamputraintan:
The comparison is broken because the tracking sheet is all string and coverage is stored in float. It will then think this is always an update although the values are the same.

Fix #690


reisingerf (approved):
Thank you!

victorskl (approved):
LGTM

PR #692, opened 2024-11-14
Added lambda layer for workflows API endpoint + RNASum Lambda

alexiswl:


victorskl (approved):
Sound good. Thanks for quick remedy.

PR #625, opened 2024-10-30
feat(event-source): add archive bucket as event source

mmalenic:
Related to https://github.com/umccr/infrastructure/pull/490

### Changes
* Adds archive bucket as event source for filemanager.

PR #700, opened 2024-11-15
feat(filemanager): add fastq bucket to ingester event source

mmalenic:
Similar to https://github.com/umccr/orcabus/pull/625

### Changes
* Adds fastq bucket as event source for filemanager.

I think this will need a change to the role similar to [this](https://github.com/umccr/infrastructure/blob/586a87926c90f9c913170096f5670105586abd4f/terraform/stacks/unimelb/data_archive/analysis_archive.tf#L76-L101)?


reisingerf (approved):

victorskl (approved):

victorskl:
> re: need a change to the role similar 

Probably. Could pls PR over there, Tq.

PR #703, opened 2024-11-17
Allow validation only for ora decompression

alexiswl:
Given a raw md5sum and a ora file, confirm that the ora file when decompressed to raw format matches the md5sum provided.  

TODO 
- [x] md5sum generation of the raw file from gz takes a lot longer (24 hours if done sequentially for every file). Working on fixing up CWL code to run md5sum in parallel


victorskl:
Related https://umccr.slack.com/archives/C025TLC7D/p1731647292332159

victorskl (approved):

PR #709, opened 2024-11-20
feat: dump postgres records

### LINKED ISSUE #705: feat: add Lambda function to dump development database records
A Lambda function should be added to dump the development database to an S3 bucket for local development.

mmalenic:
Closes #705

### Changes
* Adds `pg-dd` command that can dump the local database, upload/download dumps to S3, and load dumps into the database.
    * This is based on `copy ... from/to stdin/stdout` SQL command. The reason this is used of pg_dump is because it's easier to embed in a Lambda function, and it allows executing custom sql queries to fetch records (which is used for filemanager).
    * Using Python because `psycopg` has good support for this compared to TypeScript.
* Deploy this code as a Lambda function which can be used to dump records from the running dev RDS instance to S3.


victorskl (approved):
LGTM. 
I will give it a try soon.

mmalenic:
### Changes
* Dockerfile is simplified to call `poetry run` directly instead of trying to create a multi-stage build, in order to increase compatibility.
* The `upload` and `download` commands are more intuitive, they now only read from/write to the output directory, they don't actually connect to the database.
    * There is now an extra option for the upload subcommand `--dump-db` that allows dumping the database to the local directory first before uploading.

williamputraintan (approved):
LGTM!

mmalenic:
No idea why, but for some reason when writing the object it in `cli download`, it gets compressed twice. Fixed with recent commit.

williamputraintan:
> Fixed with recent commit.

Could confirm the fix here, thanks!

PR #730, opened 2024-11-29
Set sample type to ctDNA for cttsov2 triggers

### LINKED ISSUE #717: Only ctDNA types should go through the cttsov2 pipeline
ctTSOV2 and ctTSO are the chemistry types (ASSAY) but we may run WGS through this chemistry (for lols?), these should not go through the ctTSOV2 pipeline and should not go through PierianDx.

alexiswl:
Resolves #717


victorskl (approved):
LGTM

PR #731, opened 2024-11-29
Don't extend portal run id to sequencer run,

### LINKED ISSUE #729: Portal Run ID is duplicated on sequencer run information for PierianDx

alexiswl:
The portal run id is already part of the case accession id, which is why it is currently duplicated

Resolves #729


victorskl (approved):
LGTM

PR #726, opened 2024-11-28
feat(data-migrate): data mover task

mmalenic:
Implements first task from #721

### Changes
* Adds a data mover fargate task which can be triggered by a step function.
    * Opted for simpler 2x `aws s3 sync` + `aws s3 rm` to perform the move.
* Permissions required are potentially separate for the read from bucket and the write to bucket.

PR #735, opened 2024-11-29
fix(data-migrate): set arch as ARM64

mmalenic:
Related to #726

### Changes
* Use ARM64 instead of X86_64.


victorskl (approved):
Ah ha, right! Thanks for the quick fix. LGTM

PR #749, opened 2024-12-03
fix(data-migrate): fix stdout capture and failed step functions

mmalenic:
Related to https://github.com/umccr/orcabus/pull/726

### Changes
* Fixes stdout capture so that the task fails if the last sync contains output.
* Signals to the step function to fail the step on exceptions to avoid the step function hanging until timeout.

### DIFF
--- a/lib/workload/stateless/stacks/data-migrate/data_mover/cli.py
+++ b/lib/workload/stateless/stacks/data-migrate/data_mover/cli.py
@@ -9,6 +9,15 @@ logger = logging.getLogger()
 logger.setLevel(logging.DEBUG)
 
 
+def main():
+    try:
+        cli(standalone_mode=False)
+        sys.exit(0)
+    except Exception as e:
+        DataMover.send_failure(str(e))
+        sys.exit(1)
+
+
 @click.group()
 def cli():
     pass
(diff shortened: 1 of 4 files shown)


victorskl (approved):

reisingerf (approved):

alexiswl:
Sorry, just saw this, but you might be better off using the 'returncode' attribute of subprocess.run to determine if `aws s3 sync` has succeeded. 

See https://github.com/umccr/orcabus/blob/ed949c5280db406688242a5ee8b64d03b37f8c01/lib/workload/stateless/stacks/stacky-mcstackface/glue-constructs/clag/part_2/fastq-list-rows-event-shower/lambdas/get_fastqc_stats/run_fastqc.py#L129-L139 as an example

Also you can use `.decode()` to convert the stderr and stdout to strings from 'byte' types, warning though, the stderr and stdout attributes of the process are 'generators', so if you try and collect it a second time it will be empty.

mmalenic:
> you might be better off using the 'returncode' attribute of subprocess.run to determine if aws s3 sync has succeeded.

`check=True` should also be checking if the command succeeded right? I think I have to compare the stdout anyway because I need to check if `aws s3 sync` contains empty output vs non-empty output.

Issue #721, opened 2024-11-26
Add data mover service


mmalenic:
Create a new service which can move data from one place to another. For now, this is meant to move data based on a `portalRunId` in the cache bucket, to destination directory with the same `portalRunId` in the archive bucket. 

There should be two parts to this service:
* The actual mover task/job. This can be a fargate task that does the move. Some options for how to implement:
    * Script the AWS CLI to `sync` in Bash/Python. This is probably the simplest.
    * Use rclone?
* The event trigger/wrapper which consumes an `ArchiveData` (name pending) event and starts the mover task. There should also be an event that triggers when the archiving is complete.
    * This could be implemented as step functions, which trigger the task and wait until it is completed.
    
### Tasks
- [ ] Implement mover task/job.
- [ ] implement event trigger component.

@reisingerf @alexiswl @victorskl let me know if this sounds right or if anything should be added/changed.

ohofmann:
Also cc'ing @andrewpatto here

andrewpatto:
The Steps S3 copy service does copies of objects listed in a source CSV - fanning out jobs to Fargate with an adjustable fan size.

It uses rclone under the hood.

It's main advantage is that you can specify the full list of files in CSV of arbitrary size and it can consume using a Steps Distributed Map. So is good for selective copying. If you just want to do a `rclone src/* dest/` then possibly other options are more straightforward.

andrewpatto:
Because you specify each file in the source CSV - it can also do selective tasks on each individual file. So it can thaw files from Glacier.. or could be taught to decompress oras etc.

andrewpatto:
It is also possible that rclone is not quite right for the task - am toying with a true S3 clone CLI tool as part of GUARDIANs (and building on the cloud checksum tool). But that would be way in the future. For now rclone is fine.

mmalenic:
Happy to use steps-s3-copy for the first part - it seems like it does this and more?

> The Steps S3 copy service does copies of objects listed in a source CSV - fanning out jobs to Fargate with an adjustable fan size.

Does it also support copying all objects via a key prefix, E.g. copy from `<bucket1>/<prefix>/*` to `<bucket2>/<prefix>/`?

Also, it's just doing the copy right? If I wanted to delete objects at the source (to make it a move) that's separate?

andrewpatto:
> > The Steps S3 copy service does copies of objects listed in a source CSV - fanning out jobs to Fargate with an adjustable fan size.
> 
[...]

PR #754, opened 2024-12-05
fix(data-migrate): add structured output to step functions

mmalenic:
Related to https://github.com/umccr/orcabus/issues/721

### Changes
* Instead of trying to output the full `stdout` of `aws s3 sync` from the step functions ECS task, use a simplified structured output to avoid the payload limit error.
* Fix output logs so that they show in CloudWatch.


victorskl (approved):
LGTM

PR #755, opened 2024-12-06
Fix(SRM)  SRM api issue

raylrui:
1. add Post, Patch, Delete method in the api gateway
2. make run_folder_path nullable as ICAv2 event upgrade


raylrui:
Merge as hot fix to the SRM api, no changes to current workflows logic

PR #756, opened 2024-12-06
Fix/add delete for api allow methods

raylrui:
Fix of api call with DELETE method


raylrui:
Merge as hot fix to api gateway side, no break of any existing workflows

PR #762, opened 2024-12-09
feat: filemanager multiple same-key queries

### LINKED ISSUE #707: filemanager: API should allow singular and list-based queries for the same key
The filemanager API can use `[]` to query for parameters that should be combined with an "or" condition. Query parameters without `[]` get combined into "and" conditions.

The following should be possible:

```
curl -H "Authorization: Bearer $TOKEN" https://file.dev.umccr.org/api/v1/s3?key=*.fastq.gz&key[]=*123*&key[]=*456*
```

Which could combine into a statement that looks like:

```sql
select *
from s3_object
where key like '%.fastq.gz' and (key like '%123%' or key like '%456%')
```

Currently this is not possible and results in a deserialization error:

```
"Failed to deserialize query string. Error: Multiple values for one key: \"key\""
```

Note, that I think the order of the query parameters shouldn't matter, so if they are out of order, keys with `[]` would be grouped anyway. Fixing this still won't enable complex logic that involves nesting conditions.

mmalenic:
Closes #707 

### Changes
* Adds explicit option to specify `and` or `or` conditions for multiple keys in the API.
    * By default or is used if omitted with `key[]=...`, like before.
    * Use `key[or][]=...` to explicitly request `or` joining, or `key[and][]=...` to explicitly request `and` joining. The extra `[]` is required.
* General dependency bumps and OpenAPI generation update.


victorskl (approved):
LGTM. 
Let give this a go!

williamputraintan (approved):
LGTM

PR #765, opened 2024-12-09
Feat(MM): Additional MM Endpoint and Library Filter

### LINKED ISSUE #575: metadata-manager: Cleanup Dangling Record
The auto-sync/loader may delete libraries record if no longer exists for that particular year (as described in [how-syncing-the-data-works](https://github.com/umccr/orcabus/tree/main/lib/workload/stateless/stacks/metadata-manager#how-syncing-the-data-works) readme). We should have a way perhaps a lambda to identify and clean records that don't have a link to a library.

### LINKED ISSUE #578: Show error message in payload for icav2 workflows that fail
A bit of hacking required for cttso v2 runs (that have succeeded) but for general ICAv2 runs that have failed, we should be able to grab the 'error' message attribute from the ICAv2 event

williamputraintan:
- Add Deletion and Patch endpoint for each metadata model (JWT email token will track the user)
- Add `is_library_none` filter to Sample and Subject model to find the library-less entity

Resolve #575

### DIFF
--- a/lib/workload/components/api-gateway/index.ts
+++ b/lib/workload/components/api-gateway/index.ts
@@ -99,6 +99,7 @@ export class ApiGatewayConstruct extends Construct {
           CorsHttpMethod.OPTIONS,
           CorsHttpMethod.POST,
           CorsHttpMethod.PATCH,
+          CorsHttpMethod.DELETE,
         ],
         allowOrigins: props.corsAllowOrigins,
         maxAge: Duration.days(10),
--- a/lib/workload/stateless/stacks/metadata-manager/app/tests/test_viewsets.py
+++ b/lib/workload/stateless/stacks/metadata-manager/app/tests/test_viewsets.py
@@ -1,12 +1,17 @@
+import json
 import logging
 
 from django.test import TestCase
 
+from app.models import Library, Sample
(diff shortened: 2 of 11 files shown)


victorskl (approved):
LGTM

reisingerf (approved):

reisingerf:
The idea is to guard those endpoints with Verified Permissions?

williamputraintan:
> The idea is to guard those endpoints with Verified Permissions?

Yes, currently is using the same Amazon Verified Permission (AVP) which currently only has admin/no-admin. I suppose we would only need to add new policy at the AVP for these endpoints for more complex auth.

reisingerf:
Yup, all good!
As long as we can prevent uncontrolled access it's fine for now. 
We'll get more detailed use cases for the new policies soon enough ;-)

victorskl:
Very soon...

victorskl:
Btw, just for my understanding/context - may be you might have discussion with @alexiswl on this.

So, the issue #578 reported by Alexis related to (fixed by) this enhancement at MM, correct?

williamputraintan:
> Btw, just for my understanding/context - may be you might have discussion with @alexiswl on this.
> 
> So, the issue #578 reported by Alexis related to (fixed by) this enhancement at MM, correct?

Sorry, I link the wrong issue. It was supposed to be #575

PR #769, opened 2024-12-10
Added readme for launching the tumor normal pipeline

alexiswl:
Certainly incomplete, and need to provide an example for a library that actually exists in production


victorskl:
Early days....

PR #778, opened 2024-12-16
Remove null inputs from inputs payload

alexiswl:
This occurs when no fastqs are in ora format and therefore the ora reference is omitted. 

Created this step with JSONata.
@andrewpatto kinda annoying this JSONata wasn't introduced earlier (like six months ago, before I built around 70 different step functions). It's so much easier to handle / transform data in JSONata compared to JSONPath, AND there are online emulators available to help validate things prior to writing. 

A step function can now be JSONPath or JSONata, to help onboarding, the allow one to place a JSONata task INSIDE a JSONPath step function (and vice versa)
One downside is this compatibility between JSONPath and JSONata within AWS Step Functions is a bit flaky. 
For example the JSONata 'variables' functionality isn't parsed into downstream steps if the overall workflow is a JSONPath workflow. But ResultPath isn't supported in JSONata so it's difficult to place an output without removing the information in the previous state. 

In this case, I need to use 'output' instead of 'variables' and then append the value I want to the state input.

```
{% 
  (
    $input_json_null_removed := $each(
      $states.input.merge_boolean_enable_parameters_step.input_json, 
      function($v, $k){ 
        ($v != null) ? { $k: $v } : {} 
      }
    ) ~> $merge;
    /*
    FIXME, solution to combine jsonata to jsonpath
    Since variables are not supported in jsonpath
    And ResultPath is not supported in jsonata
    */ 
    $merge(
      [
        $states.input,
        {
          "input_json_null_removed": $input_json_null_removed
        }
      ]
    )
  )
%}
```

Read more on JSONata here: 
https://docs.aws.amazon.com/step-functions/latest/dg/workflow-variables.html

JSONata playground also very useful - https://www.stedi.com/jsonata/playground

### DIFF
--- a/lib/workload/stateless/stacks/transcriptome-pipeline-manager/step_functions_templates/set_wts_cwl_inputs_sfn.asl.json
+++ b/lib/workload/stateless/stacks/transcriptome-pipeline-manager/step_functions_templates/set_wts_cwl_inputs_sfn.asl.json
@@ -520,6 +520,12 @@
       "ResultPath": "$.merge_boolean_enable_parameters_step",
       "Next": "Update Database Entry"
     },
+    "Remove Null objects": {
+      "QueryLanguage": "JSONata",
+      "Type": "Pass",
+      "Output": "{% \n  (\n    $input_json_null_removed := $each(\n      $states.input.merge_boolean_enable_parameters_step.input_json, \n      function($v, $k){ \n        ($v != null) ? { $k: $v } : {} \n      }\n    ) ~> $merge;\n    /*\n    FIXME, solution to combine jsonata to jsonpath\n    Since variables are not supported in jsonpath\n    And ResultPath is not supported in jsonata\n    */ \n    $merge(\n      [\n        $states.input,\n        {\n          \"input_json_null_removed\": $input_json_null_removed\n        }\n      ]\n    )\n  )\n%}",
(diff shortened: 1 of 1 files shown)


alexiswl:
And the ternary operator is a massive game changer!

Syntax is

```
<BOOLEAN_EVALUTION> ? "VALUE IF TRUE" : "VALUE IF FALSE"
```

An example is

```
Price < 50 ? "Cheap" : "Expensive"
```

But can also use for a default value if a variable is undefined

```
$my_null_var ? $my_null_var : "Default value"
```

alexiswl:
Also in the command above the `~>` is the 'chain' operator https://docs.jsonata.org/other-operators#-chain

andrewpatto:
Yeah I played with it for a single steps in holmes - as exactly like you I had gone and implemented a bunch of code the old JSONpath way a week before the release..

andrewpatto:
Is there going to be a need to rewrite some of the Holmes steps to make them invokable as a sub-step?

victorskl (approved):

alexiswl:
> Is there going to be a need to rewrite some of the Holmes steps to make them invokable as a sub-step?

No, one just needs to set the `QueryLanguage' attribute inside the Task Json to 'JSONata'.

PR #783, opened 2024-12-17
fix(pg-dd): python version and environment variables

### LINKED ISSUE #782: pg-dd: service should not use python 3.13
The newest supported Python Lambda runtime is 3.12, and pg-dd uses 3.13. This causes import errors because requirements cannot be met for 3.13 packages inside the 3.12 runtime.

mmalenic:
Closes #782

### Changes
* Downgrade python version to 3.12.
* Update environment variables in Lambda function.


williamputraintan (approved):
LGTM, thanks for the fix!

victorskl (approved):

PR #784, opened 2024-12-17
Chore: Refactor Django `orcabus_id`

### LINKED ISSUE #766: Chore: Refactor OrcaBus Id for Django
Expirement and/or implement `orcabus_id` as a Django Custom ModelField. Potentially this could extract the logic of orcabus_id prefix out from the app implementation.

williamputraintan:
Implement custom Django [custom-model-field](https://docs.djangoproject.com/en/5.1/howto/custom-model-fields/) for `orcabus_id` to include prefixes.

Resolve #766 

The Django custom works well and could pass with or without the prefix. A function is called to sanitize the `orcabus_id` before or after hitting the database.


lib/workload/stateless/stacks/metadata-manager/app/models/base.py
```diff
@@ -42,7 +43,9 @@ def reduce_multi_values_qor(key: str, values: List[str]):
         ):
             values = [values]
         return reduce(
-            operator.or_, (Q(**{"%s__iexact" % key: value})
+            # Apparently the `get_prep_value` from the custom fields.py is not called prior hitting the Db but,
+            # the regular `__exact` still execute that function.
+            operator.or_, (Q(**{"%s__exact" % key: value})
```
williamputraintan, line 48: One caveat I found here is the strip/annotation will not work if it uses `iexact` on the filtering based on the field model. Currently changing the filter to only match for an exact match.

victorskl:
.. will review in tick

victorskl (approved):
LGTM. Great effort!

raylrui (approved):
Great effort!

PR #793, opened 2024-12-23
feat(filemanager): update ingest id

### LINKED ISSUE #792: filemanager: add way to manually link objects
The filemanager API should support a method to manually link objects, in case the `ingestId` automatic linking fails. This could take a similar form to the manual API calls which allow annotating objects with attributes like the `portalRunId`.

For example, we could first tag objects on the `portalRunId`, and then link all objects with the same `portalRunId`:

```
/api/v1/s3/link?attributes[portalRunId]=*20220416e17de179*
```

Care needs to be taken to ensure that the correct object gets linked with the correct id in each directory. How should this be determined? Could objects with the same `etag` and size be linked? Or the same key suffix? Otherwise objects potentially could be enumerated and linked individually:

```
/api/v1/s3/link?bucket=pipeline-prod-cache-503977275616-ap-southeast-2&key=byob-icav2/production/analysis/wts/20220416e17de179/MDX220129_somalier/MDX220129.somalier&withBucket=archive-prod-analysis-503977275616-ap-southeast-2&withKey=v1/year=2022/month=04/20220416e17de179/MDX220129_somalier/MDX220129.somalier
```

mmalenic:
Closes #792

### Changes
* Adds method to update the `ingestId` using the existing JSON patch routes. This is only allowed to update if the `ingestId` column is null, it cannot overwrite an existing `ingestId`.

### DIFF
--- a/lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md
+++ b/lib/workload/stateless/stacks/filemanager/docs/API_GUIDE.md
@@ -199,6 +199,25 @@ curl -X PATCH -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/js
 "https://file.dev.umccr.org/api/v1/s3?key=*202405212aecb782*" | jq
 ```
 
+In addition to updating attributes, the PATCH request can also be used to update the `ingestId` for records with a null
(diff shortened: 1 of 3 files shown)


victorskl (approved):
LGTM

PR #797, opened 2025-01-03
Feat: Django Deployable Swagger

williamputraintan:
Only in Django Microservice

- Move swagger local to the deployable endpoint
- Change swagger path (from `/swagger-ui`) to `/schema/swagger-ui`
- Add NoneAuth for all path prefix with `/schema` endpoint in API-GW

cc: @raylrui @mmalenic


victorskl (approved):
LGTM. 

This gonna to be the popular one, Will! One of backlog wish list of all time. Nailed it!

PR #799, opened 2025-01-06
fix(WFM): Put Back `camelCase` Serializer

williamputraintan:
The removal in #784 is not intended. The re-run workflow depends on this to dispatch the correct JSON format to the EB.

Fix [slack-thread](https://umccr.slack.com/archives/C03ABJTSN7J/p1736120499759199?thread_ts=1734914138.553039&cid=C03ABJTSN7J)


victorskl (approved):
Thanks for quick fix.

PR #800, opened 2025-01-06
Fix(WFM): Portal Run Id Creation

williamputraintan:
Fix incorrect `portal_run_id` format

PR #801, opened 2025-01-07
fix(filemanager): pin htsget-rs version inside the htsget stack

mmalenic:
### Changes
* Add the `gitReference` when fetching htsget-rs to avoid errors with versions from the main branch.

This should fix build errors in #800 /cc @williamputraintan


victorskl (approved):
LGTM

Issue #758, opened 2024-12-06
feat: add htsget support


mmalenic:
The orcabus should support using htsget-rs to fetch files. This should be enabled for all htsget-compatible files (BAM, CRAM, BCF, and VCF) that the filemanager has access to.

PR #805, opened 2025-01-09
build: update htsget-lambda dependency

mmalenic:
Related #758 
Related PR: https://github.com/umccr/htsget-deploy/pull/6

### Changes
* Update `htsget-lambda` with `aws-cdk-lib` peer dependency so that the `aws-cdk-lib` dependency can be updated in orcabus.


victorskl (approved):
LGTM. 
Thanks for quick look into this and fix, Marko.

PR #806, opened 2025-01-13
Feat(mm): add `override_cycles` to the Library

williamputraintan:
- add `override_cycles` to Library model to accommodate SSChecker

Related #798

PR #807, opened 2025-01-13
build(filemanager): set cargo-chef version to 0.1.68

mmalenic:
This should allow #806 build to succeed.

### Changes
* Temporarily set cargo-chef version in filemanager's `Dockerfile` to `0.1.68`.
    * Also adds ability to set versions for other dockerfile dependencies.
    * The latest [cargo-chef](https://github.com/LukeMathWalker/cargo-chef/issues) version seems to be broken.


williamputraintan (approved):
Thanks for the speedy fix! I haven't even reached out yet  😁

PR #813, opened 2025-01-16
fix(token-service): generated password should contain required characters

### LINKED ISSUE #811: bug: token service secret rotation sometimes fails
The token service Lambda rotation function sometimes fails with an error:

```
An error occurred (InvalidPasswordException) when calling the AdminSetUserPassword operation: Password does not conform to policy: Password must have symbol characters
```

I think the line that generates a random password in the token service won't always generate passwords that conform to the requirements, because the random generator won't necessarily include special characters. I.e. it might randomly choose non-special characters for all choices: 
https://github.com/umccr/orcabus/blob/f38855e3ed1ea270718f38c0cf5564737dac5f84/lib/workload/stateful/stacks/token-service/token_service/cognitor/__init__.py#L32

mmalenic:
Closes #811 

### Changes
* Ensure that the required characters are always present in generated password to prevent the secret rotation from failing.
    * Cognito allows defining password requirements, in this case the password should have at least 1 uppercase, lowercase, numeric, and special characters: https://docs.aws.amazon.com/cognito/latest/developerguide/managing-users-passwords.html

### DIFF
--- a/lib/workload/stateful/stacks/token-service/token_service/cognitor/__init__.py
+++ b/lib/workload/stateful/stacks/token-service/token_service/cognitor/__init__.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
-import random
+import secrets
 import string
 from dataclasses import dataclass
+from typing import List
 
 import boto3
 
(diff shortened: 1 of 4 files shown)


reisingerf (approved):
Sweet! LGTM

mmalenic:
I needed to add the ability to reset the secret if it gets stuck in the `AWSPENDING` state. It's not possible to create a new pending secret if there is already an existing pending secret, because the existing secret is always retrieved:

https://github.com/umccr/orcabus/blob/32faa5c230203e056f07697e2d8b8882ab577687/lib/workload/stateful/stacks/token-service/token_service/rotate_service_user.py#L122-L124

Now there is another step called `"resetPendingSecret"` which can be called manually.

PR #814, opened 2025-01-17
fix(pipeline): enable logging on github action runner

### LINKED ISSUE #812: bug: github action runner build project should enable logging
I think the github action runner build project should have logging enabled. However, it's been disabled explicitly - there could be a reason for this?

https://github.com/umccr/orcabus/blob/f38855e3ed1ea270718f38c0cf5564737dac5f84/lib/pipeline/statelessPipelineStack.ts#L50

mmalenic:
Closes #812 

### Changes
* Enables logging on the github action runner to address security hub control: https://docs.aws.amazon.com/securityhub/latest/userguide/codebuild-controls.html#codebuild-4

### DIFF
--- a/lib/pipeline/statelessPipelineStack.ts
+++ b/lib/pipeline/statelessPipelineStack.ts
@@ -1,9 +1,12 @@
 import { Construct } from 'constructs';
 import * as cdk from 'aws-cdk-lib';
+import { ArnFormat, Stack } from 'aws-cdk-lib';
 import * as ssm from 'aws-cdk-lib/aws-ssm';
 import * as pipelines from 'aws-cdk-lib/pipelines';
 import * as codebuild from 'aws-cdk-lib/aws-codebuild';
+import { ComputeType } from 'aws-cdk-lib/aws-codebuild';
 import * as iam from 'aws-cdk-lib/aws-iam';
+import { PolicyStatement } from 'aws-cdk-lib/aws-iam';
 import * as chatbot from 'aws-cdk-lib/aws-chatbot';
 import * as codepipeline from 'aws-cdk-lib/aws-codepipeline';
 import * as codestarnotifications from 'aws-cdk-lib/aws-codestarnotifications';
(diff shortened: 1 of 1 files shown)


lib/pipeline/statelessPipelineStack.ts
```diff
@@ -26,10 +28,26 @@ export class StatelessPipelineStack extends cdk.Stack {
...
+      resource: 'log-group',
+      arnFormat: ArnFormat.COLON_RESOURCE_NAME,
+      resourceName: `/aws/codebuild/${name}`,
+    });
+    ghRunnerRole.addToPolicy(
+      new PolicyStatement({
+        resources: [logGroupArn, `${logGroupArn}:*`],
+        actions: ['logs:CreateLogGroup', 'logs:CreateLogStream', 'logs:PutLogEvents'],
+      })
+    );
```
mmalenic, line 45: I'm not certain that this is correct, is there a way I can test it?

Otherwise, I could also just allow on all resources using `*`.

williamputraintan: I think It's about right.

williamputraintan: I suppose the `effect` here is default to allow?

williamputraintan (approved):

lib/pipeline/statelessPipelineStack.ts
mmalenic: Yeah, should be defaulting to allow, although it's undocumented: https://github.com/aws/aws-cdk/blob/53dc0d81e004ad1f8513af534b6d43f4315c24a3/packages/aws-cdk-lib/aws-iam/lib/policy-statement.ts#L90C19-L90C22

mmalenic:
I realised I can just use the L2 construct and the `WORKFLOW_JOB_QUEUED` which creates nearly identical cloudformation templates.

williamputraintan:
> I realised I can just use the L2 construct and `WORKFLOW_JOB_QUEUED` which creates nearly identical cloudformation templates.

Yes, nice! it wasn't available last time https://github.com/aws/aws-cdk/issues/30062 but they seem to add this in.

williamputraintan:
On the L2 Construct does it have a retention policy config for the logs? Or is it set by default?

mmalenic:
I've added a log group with a retention policy now. Permissions still get generated as long as the name of the log group is `/aws/codebuild/${projectName}`.

reisingerf (approved):
LGTM

PR #819, opened 2025-01-22
Added dynamodb-fastapi backed fastq manager

alexiswl:
Fastq Manager using FastAPI and Dyntastic as backends. 

Uses lambdas to communicate between the file manager and the metadata manager when required.


alexiswl:
This is working, I'd like to be able to deploy this to production so that I can start populating it and testing integration with other services such as workflow services

victorskl:
Let get this reviewed and push out, folks.!

reisingerf (approved):
Let's rock (the boat)!

alexiswl:
Waiting on https://github.com/umccr/orcabus/pull/851, I'd like to add 'storageClass' as a 'computed field' to file object readsets.

lib/workload/components/python-filemanager-tools-layer/filemanager_tools_layer/src/filemanager_tools/utils/file_helpers.py
```diff
@@ -0,0 +1,167 @@
...
+        "currentState": 'true'
+    })
+
+    if len(response) == 0:
+        # Try again with current_state=False
+        response = get_request_response_results(S3_LIST_ENDPOINT, {
+            "bucket": s3_bucket,
+            "key": s3_key,
+            "currentState": 'false'
+        })
```
mmalenic, line 31: Non-current records might not have any data on S3, except for something like old versioned objects. Just something to keep in mind, I'm not 100% sure if this is being used to presign URLs eventually.

mmalenic (approved):

victorskl:
Thanks for code walkthrough / demo today, Alexis!

alexiswl:
Suprseded by #897

PR #821, opened 2025-01-23
add lookup_value_regex to allow special char in url

raylrui:
1. lookup_value_regex = "[^/]+" # to allow id prefix like 'seq.' in the url
2. remove prefix when prefix added in "associated_id"

### DIFF
--- a/lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager/viewsets/comment.py
+++ b/lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager/viewsets/comment.py
@@ -19,7 +19,7 @@ class CommentViewSet(mixins.CreateModelMixin, mixins.UpdateModelMixin, mixins.Li
 
     def get_queryset(self):
         return Comment.objects.filter(
-            association_id=self.kwargs["orcabus_id"].split('.')[1],
+            association_id=self.kwargs["orcabus_id"],
             is_deleted=False
         )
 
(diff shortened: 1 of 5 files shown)


raylrui:
I will revert this [PR](https://github.com/umccr/orca-ui/pull/105) from UI when this PR merge to the Prod, because pord will has issues about api call without this changes here. @williamputraintan

lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager/viewsets/comment.py
```diff
@@ -40,7 +41,7 @@ def create(self, request, *args, **kwargs):
 
         # Add workflow_run_id to the request data
         mutable_data = request.data.copy()
-        mutable_data['association_id'] = seq_orcabus_id
+        mutable_data['association_id'] = seq_orcabus_id.split('.')[1] # remove prefix for association_id
```
williamputraintan, line 44: I think the `OrcaBusIdField` do this already?

williamputraintan (approved):

lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager/viewsets/comment.py
raylrui: sorry, just miss that, not work before but it works now

PR #823, opened 2025-01-23
Fix: `orcabus_id` marked required in OpenAPI schema response

williamputraintan:
Due to the limitation of the auto-generated OpenAPI schema by drf-spectacaular, fields that is marked `pk=True` or `blank=False` or `null=False` do not mark this as required in the response type when `default` is used in the field.

This will need a manual override on the serializer class.


raylrui (approved):

reisingerf (approved):
LGTM!
Seems easy enough to manage with the mixin

PR #830, opened 2025-01-29
Feat(authorization): Add curators group permissions

### LINKED ISSUE #829: Feat (authorization): Allow curators group to trigger rerun workflowrun
Add Curator group policy to allow the rerun API  from `workflow` microservice

williamputraintan:
- expand AVP schema and authorization to decide based on resource/endpoints
- add `curators` and `bioinfo` group policy to allow certain endpoint in each microservice
	- allow rerun workflow - `curators`, `bioinfo`
	- allow sync from external metadata sheet - `bioinfo`

Closes #829


reisingerf:
Looks good to me!
Sorry, a bit behind on this, but these policies only apply to the POST endpoint/methods at this time (e.g. as defined in the `action` of the policy statement), right?

All other endpoints/methods access is controlled by a default policy or are outside VerifiedPermissions e.g. no authorisation logic needed, authentication is sufficient?

reisingerf (approved):
Happy to give this a go!

williamputraintan:
> All other endpoints/methods access is controlled by a default policy or are outside VerifiedPermissions e.g. no authorisation logic needed, authentication is sufficient?

Yes, so
- all `GET` and some WFM endpoints (other than workflow rerun) are not using verified permission so it uses the default cognito JWT verification.
- Endpoints that use verified permission (metadata ext sync, wfm rerun, FM annotation endpoints) are blocked unless it is whitelisted here based on the roles (admin, curators, bioinfo). The admin (which also be the service JWT token) is given a wildcard to always permit any action

reisingerf:
Yup, all good!
Just wondering if / where we'd have the overview of which permissions / access is applied where and how. 

No worries! We'll revisit at a later stage when we have a better understanding of all this.

PR #839, opened 2025-01-30
refactor srm srv for detail states and comment

### LINKED ISSUE #848: SRM add detail state record
As status group logic for the BaseSpace_Sequence states, refer code [here](https://github.com/umccr/orcabus/blob/516e94e245bb3d902a95d6abada08a76c2e5e0db/lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager/models/sequence.py#L31C5-L31C41), some state changes have been ignore in state record service.

raylrui:
Sequence Run Manager service refactor

1. add detail state record for srm_srv states
2. add comment for "Sequence completed", "Sequence completed. But failed in post analysis process.", and "Sequence failed."

fix #848

### DIFF
--- a/lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager_proc/services/sequence_state_srv.py
+++ b/lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager_proc/services/sequence_state_srv.py
@@ -32,24 +32,4 @@ def create_sequence_state_from_bssh_event(payload: dict) -> None:
     # None by default
     comment = None
     
-    # sequence status in SUCCEEDED: 
-    # if transition from STARTED to SUCCEEDED, set comment to "Sequence complete"
-    # if transition from FAILED to SUCCEEDED, set comment to "Conversion re-triggered, and sequence completed.'
-    if sequence.status == SequenceStatus.SUCCEEDED:
-        previous_state = State.objects.filter(sequence=sequence).order_by('-timestamp').first()
-        if previous_state is not None and SequenceStatus.from_seq_run_status(previous_state.status) == SequenceStatus.STARTED:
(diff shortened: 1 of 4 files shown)


lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager_proc/services/sequence_state_srv.py
```diff
@@ -29,7 +29,27 @@ def create_sequence_state_from_bssh_event(payload: dict) -> None:
...
+    # sequence status in FAILED:
+    # if transition from SUCCEEDED to FAILED, set comment to "Sequence completed. But failed in post analysis process."
+    # if transition from STARTED to FAILED, set comment to "Sequence failed. Now in state " + status + " ."
+    if sequence.status == SequenceStatus.FAILED:
+        previous_state = State.objects.filter(sequence=sequence).order_by('-timestamp').first()
+        if previous_state is not None and SequenceStatus.from_seq_run_status(previous_state.status) == SequenceStatus.SUCCEEDED:
+            comment = "Sequence completed. But failed in post analysis process."
+        else:
+            comment = "Sequence failed. Now in state " + status + " ."
+
```
reisingerf, line 54: All this is to set a comment? 

I don't think that's needed. With the separation of state and status, it should be quite clear what's going on. 
I'd preserve the comments for irregular, unpredictable, manual user added text rather than being auto-populated.
(if something is predictable and can be auto-generated I always feel it should be in a structured format and not in free text, and with the state / status we have that already)

raylrui: yes, sure, these can be removed as redundant for states.

reisingerf (approved):

PR #840, opened 2025-01-30
fix(pg-dd): remove double compression

### LINKED ISSUE #838: pg-dd: don't compress twice
Some parts of the code pg-dd code compress files twice, once when writing to a local file, and then again when uploading to S3. This causes code that reads these files to fail.

mmalenic:
Closes #838 

### Changes
* Avoid compressing and decompressing in `PgDDS3`, and only compress/decompress in `PgDDLocal`.
* Also updates lock file.


raylrui (approved):
LGTM

PR #843, opened 2025-01-31
SRM add time range filter and stats count api

raylrui:
api support to issue: https://github.com/umccr/orca-ui/issues/103

1. add support for time range selection:
`/api/v1/sequence/?page=1&rowsPerPage=10&status=failed&start_time=2024-10-29T00:00:00%2B00:00&end_time=2024-10-30T00:00:00%2B00:00`
2. add status count api for status selection
image


raylrui:
merge, just add new filter api for SRM

PR #850, opened 2025-02-04
Feat(WFM): Duplication rerun check

### LINKED ISSUE #815: WorkflowManager: extend workflow re-run condition check
When a workflow is retriggered via the API / UI the default behaviour should be to prevent reruns if the same request already exists, i.e. if the exact workflow inputs where already used in an earlier (on ongoing) workflow execution. 

This is to prevent accidentally duplicated workflow run requests. Legitimate re-runs with the exact same input should still be possible, but should require additional actions or admin intervention.

williamputraintan:
Add duplication input check before executing emitting rerun events to EB. The check is based on the `READY` state payload, where the workflow entity and linked library are the same as the given wfl_run. The payload check will depend on each wfl type, whereas RNAsum is only based on dataset input.

This will throw an error on UI as the bad request is returned.
Screenshot 2025-02-04 at 11 17 48

closes #815

### DIFF
--- a/lib/workload/stateless/stacks/workflow-manager/workflow_manager/errors.py
+++ b/lib/workload/stateless/stacks/workflow-manager/workflow_manager/errors.py
@@ -1,0 +1,8 @@
+class RerunDuplicationError(Exception):
+    """
+    Exception raised when a rerun duplication is not allowed.
+    """
+
+    def __init__(self, message="Duplication rerun is not allowed."):
+        self.message = message
+        super().__init__(self.message)
--- a/lib/workload/stateless/stacks/workflow-manager/workflow_manager/serializers/workflow_run_action.py
+++ b/lib/workload/stateless/stacks/workflow-manager/workflow_manager/serializers/workflow_run_action.py
@@ -15,6 +15,9 @@ class AllowedRerunWorkflowSerializer(serializers.Serializer):
 
 
 class BaseRerunInputSerializer(serializers.Serializer):
+    # Reruns with the same input (duplicate reruns) should not be allowed unless this flag is set to True
+    allow_duplication = serializers.BooleanField(default=False, required=False,
+                                                 help_text='Allow rerun with the same previous input (duplicate rerun)')
 
     def update(self, instance, validated_data):
         pass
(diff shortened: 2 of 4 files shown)


raylrui (approved):
LGTM.
but I am thinking, like add a new comment record for every successful rerun maybe also help for users, like this after checking the details
```
# Create rerun comment
    comment_text = (
                    f"Workflow rerun triggered by {request.user.email}\n"
                    f"Dataset: {serializer.validated_data.get('dataset')}\n"
                    f"Allow duplication: {serializer.validated_data.get('allow_duplication', False)}\n"
                    f"New Portal Run ID: {detail.get('portalRunId')}"
                )
                
     WorkflowRunComment.objects.create(
                    workflow_run=wfl_run,
                    comment=comment_text,
                    created_by=request.user.email,  # Assuming user email is available
                )
```

reisingerf (approved):

williamputraintan:
> LGTM. but I am thinking, like add a new comment record for every successful rerun maybe also help for users, like this after checking the details
> 
> ```
> # Create rerun comment
>     comment_text = (
>                     f"Workflow rerun triggered by {request.user.email}\n"
>                     f"Dataset: {serializer.validated_data.get('dataset')}\n"
>                     f"Allow duplication: {serializer.validated_data.get('allow_duplication', False)}\n"
>                     f"New Portal Run ID: {detail.get('portalRunId')}"
>                 )
>                 
>      WorkflowRunComment.objects.create(
>                     workflow_run=wfl_run,
>                     comment=comment_text,
>                     created_by=request.user.email,  # Assuming user email is available
>                 )
> ```

Hmm, I think we should have a proper auditing (of who does what) schema instead of having it as a comment. I could put this as logging here instead of persisting it in the db. I think the linking of the new and old runs should be quite straightforward from the library filtered workflow.

williamputraintan:
>  I could put this as logging here instead of persisting it in the db. 

Maybe I'll do this when we decide to have the auditing schema, it seems that we currently don't log the API requests in Django.

raylrui:
> > I could put this as logging here instead of persisting it in the db.
> 
> Maybe I'll do this when we decide to have the auditing schema, it seems that we currently don't log the API requests in Django.

sure, I am also think have a general audit system for FE, as now our UI have many interactive actions. Issue for tracking: https://github.com/umccr/orca-ui/issues/85

PR #710, opened 2024-11-21
addd basic stack cdk code

raylrui:

PR #859, opened 2025-02-12
feat: filemanager crawl

### LINKED ISSUE #852: filemanager: add crawl function
I think the filemanager API should have a crawl function. This could be based on the existing `inventory` function, but I think it might be more convenient to just use list operations directly. Often records become out of date due to missing features or bugs, so along with ingesting new objects, the crawl function could be used to correct records if they are inconsistent. By placing this inside the filemanager directly, S3 API code can be re-used, rather than tying the logic to update scripts and individual API calls.

Related to #844.

mmalenic:
Closes #852

### Changes
* Adds crawl API for sync/async crawls that tracks status in a new `s3_crawl` table in the database.
* Crawls will list all objects in the bucket and ingest records based on that, updating record metadata and tags.
* This PR also contains some cleanup of mockall `#[double]` usage, which was conflicting with `aws-smithy-mocks-experimental`

The async API isn't working yet because `HttpApi` doesn't support async lambda invocations. I'm happy with the PR for now and I'll add that in another PR.

I can either change the API to a `RestApi` or re-invoke the Lambda function from within the existing `HttpApi` endpoint and detach the execution. I'm not sure if there is a preference to use `RestApi` with any of the other microservices?


victorskl:
.. will review soon in arvo!

brainstorm (approved):

victorskl (approved):
LGTM

victorskl:
> I'm not sure if there is a preference to use RestApi with any of the other microservices?

Give it try RestApi if you really need to. Do watch out on cognito authorizer but this might work nowaday. Might hit and loose built-in JWT auth capability but, could workaround with Lambda Auth (but then ... umm, a bit less in favour as the overhead it adds, tbh).

victorskl:
Or, check with @raylrui on WebSocket - ditto #710 Fancy!

Issue #884, opened 2025-02-28
Sequence Run Manager: replace 'instrument_run_id' with 'id' for unique identifier


raylrui:
When we re-trigger a sequence run in Sequence Hub (Base Space) with same sequencing data (RunInfo.xml ), it may has same `instrumentRunId` (instrumentRunId is set by the RunInfo.xml). 
This may cause combination of different runs, and as the run `id` and `apiUrl` is not updated, the run details and sample sheet api call will be failed.

See example in dev: https://orcaui.dev.umccr.org/runs/sequence/seq.01JN2WW3887X8BH3P6NMPWVPRD
events compared:
![Image](https://github.com/user-attachments/assets/15520182-9111-4208-81a8-22dba882a1a3)

So rather than using `instrumentRunId` as the unique identifier, run 'id' should be a better choice. And in early stage of bssh events when run files are still uploading, there is no `instrumentRunId` set.

![Image](https://github.com/user-attachments/assets/4f0fb10b-a622-4dc1-9688-d76be7941d48)

raylrui:
This is now has been resolved as #885

PR #885, opened 2025-03-03
Fix(SRM)/replace 'instrument_run_id' with 'id' for unique identifier

raylrui:
Solve: https://github.com/umccr/orcabus/issues/884

1. For each 'bssh run', 'run id' is the unique identifier rather than 'instrument_run_id'. Reason explian in [884](https://github.com/umccr/orcabus/issues/884)
2. update the service to handle new changes and also the test cases


lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager/models/sequence.py
```diff
@@ -76,40 +76,27 @@ class Meta:
...
+    v1pre3_id = models.CharField(max_length=255, null=True, blank=True)
+    ica_project_id = models.CharField(max_length=255, null=True, blank=True)
+    api_url = models.TextField(null=True, blank=True)
     end_time = models.DateTimeField(null=True, blank=True)
+    run_volume_name = models.TextField(null=False, blank=False)  # legacy `gds_volume_name`
+    run_folder_path = models.TextField(null=True, blank=True)  # legacy `gds_folder_path`, nullable as ICAv2 event upgrade
+    run_data_uri = models.TextField(null=False, blank=False)  # must be absolute path, including URI scheme/protocol
 
     # optional fields -- business look up keys
+    instrument_run_id = models.CharField(max_length=255, null=True, blank=True)
```
victorskl, line 94: Pivoting to BSSH upload id (aka `run_id`) for event level record capture is fine. Understood.
Please make sure @reisingerf is okay with this changes.

One minor request though, could you mark this `instrument_run_id` field not nullable pls.

victorskl (approved):
LGTM. Minor request to attend.

lib/workload/stateless/stacks/sequence-run-manager/sequence_run_manager/models/sequence.py
reisingerf: Yes, I think I'm OK with this. 

The `run_id` is generated by BSSH upon receiving a new upload and it's the ID used internally by BSSH, e.g. in the `apiUrl`, right? 
So it represents the uniqueness identifier BSSH uses to reference run records/uploads. In that case I think it makes a good and sensible ID for the sequence manager. 

Also, agreed that we should insist on the `instrument_run_id`. I think we always have one (if the upload happens from the instruments) and we **should** provide one for manual uploads. 
(not quite sure what happens if we manually upload without providing one... @victorskl do you know? )

victorskl: This gets tricky.

Though, the `RunInfo.xml` at the data location (if we are parsing that file yet) should always contain `instrument_run_id` regardless of BSSH event. In past, we used to break it when that happen, and then we follow up as override SOP, etc. (Hence, requesting `instrument_run_id` to non null field to break it above^^.)

If we do not like to break it, perhaps, I can advice we can split the model into `Sequence` and `BSSHRunEvent`. And, establish the relation Many-to-One between event and the actual sequencing info. This might need re-work on modelling a bit to what Ray had done in recent PRs, etc.

raylrui: >One minor request though, could you mark this instrument_run_id field not nullable pls.

ah, yes, the reason I make it nullable before is that there is a `Uploading` events at very early stage like this:
image


 It has no `instrument_run_id` info ( I think it is still uploading `RunInfo.xml`), So I am think if we can capture this events as well
But I found we have events rule like this:
image

so this events actually is no captured, as no 'instrument_run_id' existed.

But it should be okay, because when we testing in the 'Uploading ' state, there is also no much information in its detail 'apiUrl', so can ignore this state.

reisingerf: Hm, I don't think we want to ignore this state... It's the earliest event we receive, right? So it's a valuable feedback.

There are other means to work around the nullable... perhaps a quick chat would be good to discuss options... 
We can then record the consensus here.

alexiswl:
Sorry, pretty late to the party on this, the v1pre3Id might be worth capturing, this is unique for every 'autolaunch' / 'run reprocess' and can be found in the output path in the ICAv2/BSSH folder. I believe it is also countable so the latest v1pre3id for a given instrument run id should be the latest 'reprocess' of a run.

raylrui:
> Sorry, pretty late to the party on this, the v1pre3Id might be worth capturing, this is unique for every 'autolaunch' / 'run reprocess' and can be found in the output path in the ICAv2/BSSH folder. I believe it is also countable so the latest v1pre3id for a given instrument run id should be the latest 'reprocess' of a run.

yes, 'v1pre3Id' already captured in our `sequence` model.
After some discussion, we prepare to make some refactor, like make `instrument_run_id` nullable to include all bssh events states, and add a new sample sheet model to preserve bssh sample sheet....
image

reisingerf:
@alexiswl thanks for the feedback! We are relying quite a bit on you here, especially in terms of what data is needed/useful to capture.
For example: we also thought that capturing the `RunInfo.xml` would be good, perhaps even the `RunParameters.xml`.
We could store the full files, analogous to what Ray is doing with the SampleSheet, but can also just retrieve some values. 
By default I'd err on the side of caution and capture the full file(s), but happy to change that if that's not useful.

alexiswl:
> For example: we also thought that capturing the RunInfo.xml would be good, perhaps even the RunParameters.xml.
We could store the full files, analogous to what Ray is doing with the SampleSheet, but can also just retrieve some values.
By default I'd err on the side of caution and capture the full file(s), but happy to change that if that's not useful.

Sorry for the delay on this, yes the v2-samplesheet-maker CLI can also parse in the run info as well and record as a JSON object

alexiswl (approved):

reisingerf (approved):

Issue #888, opened 2025-03-05
Sequence run manager: script (lambda) for creating library linking for previous sequence runs


raylrui:
# Title: 
Implement Historical Library Linking for Existing Sequence Runs

## Description
Currently, our sequence run manager contains numerous historical runs that lack library linking information. We need to implement a solution to retroactively create these missing library links.

## Proposed Solution
Create a new Lambda function that will:
1. Scan all existing sequence runs in the database
2. Identify runs without library linking
3. Build and establish library links for these historical records

## Technical Details
- Create a new Lambda function for this one-time operation
- Query existing sequence runs from the database
- Implement logic to check for missing library links
- Apply the same library linking logic used in current runs
- Update the database with the new linking information

## Acceptance Criteria
- [ ] Lambda function successfully scans all historical sequence runs
- [ ] Missing library links are identified and logged
- [ ] Library links are created for runs that lack them
- [ ] Process includes error handling and logging
- [ ] Documentation of the process and any manual steps required

## Additional Notes
- Consider implementing this as a one-time execution script
- Include monitoring and logging to track progress
- Consider rate limiting or batching to prevent database overload

PR #889, opened 2025-03-05
Library and Samplesheet Linking for Existing Sequence Runs

raylrui:
Solve: https://github.com/umccr/orcabus/issues/888

Current Process for processing library linking for existing sequence runs:
1. upload library linking files and samplesheet linking files into the temp bucket created for sequence enrichment. (linking files will have the orcabus_id/instrument_id, library collections/samplesheet json mappping)
2. trigger the lambda function with target file(key) name as input. It will the file and load the record into database by batch insertion.


victorskl (approved):
LGTM.

Minor comment, pls unit test some of these later. Okay for now as first pass.

raylrui:
>Minor comment, pls unit test some of these later. Okay for now as first pass.

sure, I will add some. And Couple of testing in dev have already conducted, it working fine. It will won't make any breaking changes as it just has insert operation. And I think we can remove it after all history record building finishing.

reisingerf (approved):

PR #890, opened 2025-03-06
feat(sscheck): Sample_Name Check

### LINKED ISSUE #883: feat(sscheck): Better Error Response
- Sync library_id extraction for each row of the sample sheet. The checking part parses the `Sample_ID` column (which is {sampleID}_{library_ID}), but the conversion to V2 takes the libraryID from the `Sample_Name` column. 
- Perhaps a more appropriate error message should also be thrown if the `Sample_Name` is not matched with `Sample_ID`

Ref:
https://umccr.slack.com/archives/C94CQMNVA/p1740625700956749

williamputraintan:
Resolve #883

### DIFF
--- a/lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/handler.py
+++ b/lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/handler.py
@@ -49,11 +49,18 @@ def lambda_handler(event, context):
                               v2_sample_sheet='')
         response = construct_response(status_code=400, body=body, origin=origin)
         return response
-
     multipart_content = {}
     for part in msg.get_payload():
         multipart_content[part.get_param(
             'name', header='content-disposition')] = part.get_payload(decode=True)
+
+    key_list = list(multipart_content.keys())
+    if "file"not in key_list or "logLevel" not in key_list:
+
+        body = construct_body(check_status="FAIL", error_message="file or logLevel not found",
+                              v2_sample_sheet='')
+        response = construct_response(status_code=400, body=body, origin=origin)
+        return response
 
     file_data = multipart_content["file"]
     log_level = multipart_content["logLevel"].decode("utf-8")
--- a/lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/src/samplesheet.py
+++ b/lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/src/samplesheet.py
@@ -341,6 +341,12 @@ class SampleSheet:
             index2 = sample_row["index2"] if "index2" in sample_row.keys() else None
             project = sample_row["Sample_Project"] if "Sample_Project" in sample_row.keys() else None
 
+            if not sample_row["Sample_ID"].endswith(sample_row["Sample_Name"]):
+                logger.error(
+                    f"Sample_Name ({sample_row["Sample_Name"]}) is not the libraryID defined in the "
+                    f"Sample_ID ({sample_row["Sample_ID"]}) format")
+                raise SampleNameFormatError
+
             self.samples.append(
                 Sample(
                     lane=lane,
--- a/lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/README.md
+++ b/lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/README.md
@@ -33,6 +33,13 @@ options:
 
 ```
 
+Set Environment Variable
+
+```shell
+export JWT_AUTH={JWT_TOKEN_HERE}
+export METADATA_DOMAIN_NAME=metadata.dev.umccr.org
+```
+
 Running example
 
 ```shell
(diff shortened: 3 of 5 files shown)


lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/src/samplesheet.py
```diff
@@ -341,6 +341,12 @@ def convert_data_to_samples(self):
             index2 = sample_row["index2"] if "index2" in sample_row.keys() else None
             project = sample_row["Sample_Project"] if "Sample_Project" in sample_row.keys() else None
 
+            if not sample_row["Sample_ID"].endswith(sample_row["Sample_Name"]):
```
williamputraintan, line 344: Mainly this, where it checks if the Sample_Name (which should be the library_id) is the same Sample_ID format (which is `{sampleID}_{library_ID}`)

lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/src/samplesheet.py
```diff
@@ -108,6 +108,12 @@ def check_sample_name_format(self):
             logger.error("Sample ID {} did not have corresponding Sample_Name".format(self.sample_id))
             raise SampleNameFormatError
 
+        # Check if the Sample_Name column is the library_id from the Sample_ID column ignoring the "_topup"
```
williamputraintan, line 111: Mainly this, where it checks if the Sample_Name (which should be the library_id) is the same Sample_ID format (which is `{sampleID}_{library_ID}`)

lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/handler.py
```diff
@@ -49,12 +49,19 @@ def lambda_handler(event, context):
...
         response = construct_response(status_code=400, body=body, origin=origin)
         return response
-
     multipart_content = {}
     for part in msg.get_payload():
         multipart_content[part.get_param(
             'name', header='content-disposition')] = part.get_payload(decode=True)
 
+    key_list = list(multipart_content.keys())
+    if "file"not in key_list or "logLevel" not in key_list:
```
alexiswl, line 58: Should be a space between 'file' and 'not'

lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/src/samplesheet.py
```diff
@@ -108,6 +108,14 @@ def check_sample_name_format(self):
             logger.error("Sample ID {} did not have corresponding Sample_Name".format(self.sample_id))
             raise SampleNameFormatError
 
+        # Check if the Sample_Name column is the library_id from the Sample_ID column ignoring the "_topup" / "_rerun"
+        clean_unique_id = SAMPLE_REGEX_OBJS["topup"].sub('', self.unique_id)
```
alexiswl, line 112: I think there's a better way to regex this, will have a closer look in the morning.

lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/handler.py
williamputraintan: Updated, although not sure why PyCharm didn't complain this.

lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/tests/sample/mock-2.csv
```diff
@@ -19,5 +19,5 @@ AdapterRead2,AGATCGG,,,,,,,,,,
 ,,,,,,,,,,,
 [Data],,,,,,,,,,,
 Lane,Sample_ID,Sample_Name,Sample_Plate,Sample_Well,Index_Plate_Well,I7_Index_ID,index,I5_Index_ID,index2,Sample_Project,Description
-3,MDX200001_L2000001,L0000000_topup,,,ABCD,ABCD,ABCD,ABCD,ABCD,ABCD,ABCD
-1,MDX200001_L2000002,L0000000_topup,,,ABCD,ABCD,ABCD,ABCD,ABCD,ABCD,ABCD
+3,MDX200001_L2000001,L2000001_topup,,,ABCD,ABCD,ABCD,ABCD,ABCD,ABCD,ABCD
```
alexiswl, line 22: So to clarify 

This would be okay

```
MDX200001_L2000001_topup,L2000001_topup
```

But this is not okay?

```
MDX200001_L2000001,L2000001_topup

```

williamputraintan: It might be me putting the test data inconsistently, but if `MDX200001_L2000001_topup,L2000001_topup` is what we aiming for, I'll add a patch

lib/workload/stateless/stacks/sample-sheet-check/sample-sheet-check-lambda/src/samplesheet.py
alexiswl: Actually because I used the non-capturing group syntax in almost all of the regexes `(?:..)` this is probably the most appropriate way forward

alexiswl (approved):

williamputraintan:
Thanks for reviewing! Merging ...

PR #897, opened 2025-03-12
Added fastq manager

### LINKED ISSUE #561: Add FastqManager
Add a service that keeps track of and manages FASTQ pairs.

alexiswl:
* Comprises fastq objects and linked fastq objects through fastq sets
* Interacts with filemanager to store ingest ids for fastq objects
* Runs QC analysis on samples with Sequali
* Runs fileCompression information allowing easy transition between ORA and GZIP
* Runs NTSM services for easy comparison within a library and for comparison between libraries

Currently running at https://fastq.dev.umccr.org/schema/swagger-ui

Supersedes #819 

Resolves #561 

# TODO
- [ ] #896


lib/workload/stateless/stacks/fastq-manager/Readme.md
```diff
@@ -0,0 +1,416 @@
...
+
+Examples of events generated are:
+
+### Fastq Created Events
+
+```json5
+{
+  "EventBusName": "OrcaBusMain",  // Name of our event bus
+  "Source": "fastq.manager",  // lowercase name of the service
+  "DetailType": "FastqListRowCreated",  // PascalCase name of the event
```
victorskl, line 45: Much better event name, good job!

Just minor note on source field, the convention is `orcabus.fastqmanager`

victorskl (approved):

reisingerf (approved):

alexiswl:
@mmalenic eventSourceConstruct tests failing at https://github.com/umccr/orcabus/actions/runs/14192658946/job/39760644268?pr=897#step:5:10

Not sure why, I haven't edited any of the event patterns or test

alexiswl:
@mmalenic I tried putting ntsm at the bottom of the list both in shared.ts and in the filemanger.ts as well (after the if conditions). 

In the tests ran a `console.log` on the event and pattern right before the line 
`expect(await testEventPattern(event, pattern)).toBe(false);`

And got the following (event)

```ts
{
        version: '0',
        id: '17793124-05d4-b198-2fde-7ededc63b103',
        'detail-type': 'Object Created',
        source: 'aws.s3',
        account: '111122223333',
        time: '2021-11-12T00:00:00Z',
        region: 'ca-central-1',
        resources: [ 'arn:aws:s3:::org.umccr.data.oncoanalyser' ],
        detail: {
          version: '0',
          bucket: { name: 'org.umccr.data.oncoanalyser' },
          object: {
            key: 'example-key/',
            size: 0,
            etag: 'b1946ac92492d2347c6235b4d2611184',
            'version-id': 'IYV3p45BT0ac8hjHg1houSdS1a.Mro8e',
            sequencer: '617f08299329d189'
          },
          'request-id': 'N4N7GDK58NMKJ12R',
          requester: '123456789012',
          'source-ip-address': '1.2.3.4',
          reason: 'PutObject'
        }
      }
```

With (pattern)

```
{
        source: [ 'aws.s3' ],
        'detail-type': [
          'Object Created',
          'Object Deleted',
          'Object Restore Completed',
          'Object Restore Expired',
          'Object Storage Class Changed',
          'Object Access Tier Changed'
        ],
        detail: { bucket: { name: [Array] }, object: { '$or': [Array] } }
      }
```

Neither of these seem out of the ordinary, but then testEventPattern fails with `Region is missing`

mmalenic:
That looks right. Do you have valid AWS credentials in your shell? This test needs access credentials as it uses the [`TestEventPattern`](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_TestEventPattern.html) API call on AWS.

There's no "local" way to test the event pattern - although there definitely should be (not sure why this is an API call).

mmalenic:
If this is something that's annoying I think this test should be ignored by default. I don't think credentials should be required just to run CDK tests.

I guess this is more of an integration test and probably belongs somewhere else.

alexiswl:
> That looks right. Do you have valid AWS credentials in your shell? This test needs access credentials as it uses the [TestEventPattern](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_TestEventPattern.html) API call on AWS.

Okay very strange, tried on my pi5 with exported AWS creds from my ubuntu tower and I had the same result. Added in the environment variables AWS_REGION (and AWS_DEFAULT_REGION) and it works now! Because the event has the 'region' key I think this made the context of the error more confusing than what it was!

PR #898, opened 2025-03-12
Add fastq unarchiving service

### LINKED ISSUE #880: Create unarchiving fastq manager
Makes a REST API endpoint to call to where a service can send through a list of fastq list row ids this will launch a job to restore this data into the 'restore' directory. 

Will have the following API endpoints

POST api/v1/job - Given a list of fastq list row ids, queries the fastq manager, generates an internal job id, kicks off the aws step function and returns a mapping to the internal job ID to the AWS step function execution id. STATUS will be RUNNING by default. Fastq list row ids returns as part of the job object. In future this will fail IF one of the following conditions is met:
  * An fqlr is part of an existing running job
  * An fqlr status is already not in deep archive - up to the service running the unarchiving step to only provide archived fqlrs

GET api/v1/job - List all jobs (by default RUNNING only)

GET api/v1/job/{job_id}- Given a job id, return the job object including status.

Will print RUNNING events when unarchiving service has started and SUCCEEDED events when unarchiving service has completed.

Services such as the fastq-glue service can listen to succeeded events and find fqlr ids in a succeeded event.

alexiswl:
Resolves #880 

Initial commit, still need to connect API and step function services. 
Will generate events on Job starts and completions. 

# TODO
- #897


victorskl:
I will review when PR is ready, Alexis.

alexiswl:
This is ready for review. 

Fastq Unarchiving endpoint can be viewed here - https://fastq-unarchiving.dev.umccr.org/schema/swagger-ui#/

Step function in dev can bee seen here - https://843407916570.ap-southeast-2.console.aws.amazon.com/states/home?region=ap-southeast-2#/statemachines/view/arn:aws:states:ap-southeast-2:843407916570:stateMachine:fastq-unarchiving-sfn

alexiswl:
Still need to wait for resolution in https://github.com/umccr/steps-s3-copy/issues/15 before I can perform a full e2e test in development. Alternatively I can push to staging or production manually along with fastq manager stack and test from there

victorskl (approved):

PR #899, opened 2025-03-12
Added fastq glue services

### LINKED ISSUE #870: Add 'fastq-glue' service
Fastq Manager shouldn't be responsible for listening to events that could be used to import fastq metadata.
However no other service should also be responsible for registering these fastqs. 

Instead we have a simple 'fastq-glue' service that performs the following:

1. Listens to the sequencer run manager event (which ever one says that a samplesheet exists) - this is hopefully before the run finishes
    * Creates fastq-sets for each library in the samplesheet, and deals with the logic of if it should be added to the existing fastq set for that library or register a new fastq set (handle topup / rerun scenario).
    * Generates an event saying 'new fastqs registered' which simply prints an instrument run id. 
      * Future glue services will be able to then subscribe to this event and query the fastq manager and decide if they need to unarchive any complementary data, and call the unarchive service, see #871, say new tumor has arrived for a subject that ran multiple months ago. Note that this would run asynchronously. Glue services instead call the fastq-sync service in #871. 

2. Listens to the future bclconvert / ora-compression pipeline that places ora files in the primary directory. 
    * Adds 'readsets' to the fastq objects mentioned above, so now fastq objects have file-information. 
    * Generates an event to say that fastq object readsets have been added to the database for the instrument run id. 
  	    * This will be the trigger for future glue services to generate 'READY' events.  

This will replace the 'clag' services of the stacky stack. No more showers :(

alexiswl:
Just added readme so far

# Fastq Glue

The fastq glue service is a simple tool that connects the fastq manager to the automation pipeline processes. 

## Starting BCLConvert

The first interaction the fastq glue service has with the automation is when the sequence run manager registers a samplesheet.  

The fastq glue service then collects the samplesheet and generates fastq sets and fastq list rows for each sample in the samplesheet.

The fastq glue service then also creates a final event with 'FastqsCreated' with an instrument run id in the detail.

Workflow glue services can then query the fastq manager for any fastq sets of interest an 'initiate' the fastq process 

## Post BSSH Fastq Copy

Once the bssh fastq copy service has completed and the fastq data is now in our BYOB bucket, the fastq glue service then performs the following.

1. Adds read sets to the fastq list rows by querying the fastq_list.csv in the instrument run id output directory.
2. Triggers qc processes endpoint for each fastq list row in the instrument run. 
3. Triggers the file compression information endpoint for each fastq list row.
4. Triggers the ntsm information endpoint for each fastq list row

Resolves #870 

## TODO

- [ ] #896 
- [ ] #897
- [ ] BSSH Fastq Copy updates with the new SRM


victorskl (approved):

reisingerf (approved):

alexiswl:
Sorry, this one is still in draft mode, will re-request review once PR is ready

alexiswl:
Superceded by #942

PR #900, opened 2025-03-12
Added Fastq Sync Service

### LINKED ISSUE #871: Add fastq-sync service
This is a ~~simple~~ service that performs the following

1. Listens to 'CheckFastqAvailableSync' events generated by workflow glue services.
    * The input contains a library id and a task token.
    * Registers the task token in the database with the library id as the primary key (as there may be multiple tasks relying on the same library id being available)
    * Workflow Glue services will not move off their step function state until they are unlocked by this service. 

2. Listens to 'FastqAvailable' events from the (future) fastq-glue services. See #870 
    * When a fastq set becomes available for that library (and has readSet information for all fastq objects in the set), then a 'sendtasksuccess' is sent through for that task token id, see point 1 above. 

3. Listens to events from glue services requesting fastq unarchiving and calls the unarchive manager (asynchronous call from glue services- see #870)

4. Listens to the unarchive manager (#869) and checks with any FastqAvailableSync pending tokens to determine if the data is available yet for their workflow.

This allows the glue services to perform the following:

1. As soon as the fastq-glue services says that there are new sets available on the instrument run, we can trigger a step function to run that starts with 'CheckFastqAvailable' events for each library required in the analysis, these will hang until the data for these services is available. 

2. Glue services can then re-query the readset now with the fastq manager now pointing to the restored file uris, and as such a READY event knowing that the data is readily available.

alexiswl:
Just added the readme for now

# Fastq Sync Service

The fastq sync service is a simple service that allows step functions with task tokens to 'hang' 
until the requirements of either a fastq list row or fastq set have been met. 

This is useful for workflow-glue services that have fastq set ids but need to wait for either

1. The fastq set readsets to be created
2. The fastq set to have been qc'd AND have a fingerprint file and compression information
3. This is also useful for data sharing services that require the fastqs to be unarchived before they can be shared

The step function will then hang at that step until the task token has been 'unlocked'

## Registering task tokens

Workflow glue services can use the fastq sync service by generating the following event

```json5
{
  "EventBusName": "OrcaBusMain",
  "Source": "doesnt matter",
  "DetailType": "FastqSync",
  "Detail": {
    "taskToken":  "uuid",
    "fastqSetId": "fqs.123456",
    // Then one or more of the following
    // Requirements can be left out if not needed
    // Do all fastq list rows in the set contain readsets?
    "hasReadsets": true,  
    // Are the fastqs required to be in active storage
    "inActiveStorage": true,
    // Do all fastq list rows in the set contain an ntsm uri?
    "hasFingerprints": true,  
    // Do all fastq list rows in the set contain compression information?
    // Useful if the fastq list rows are in ora format. 
    // Some pipelines require the gzip file size in bytes in order 
    // to stream the gzip file from ora back into s3 
    "hasCompressionInformation": true,  
    // Do all fastq list rows in the set contain qc information?
    // We don't use this for anything yet but we may use this in the future
    // to ensure that a fastq set has met the ideal coverage levels
    "hasQc": true,
  }
}
```

The fastq sync service will also trigger the qc, fingerprint or compression information services on the fastq manager if they do not exist but are required (note that the fastq sync service is aware that archived files are not able to have fastq manager internal services run on them, and as such will trigger the unarchiving of these fastqs). 

image

## Unlocking task tokens

The fastq sync service will listen for the following events:

1. FastqSetUpdated (from the fastq management service)

2. UnarchivingCompleted (from the fastq unarchiving service)

image

The fastq sync service will then check against the requirements of the fastq set or fastq list row for each task token and if so, unlock the task token.

## Launching requirements sfn

Both step functions above call the launch requirements step function on each fastq list row in the set. 

The launch requirements step function looks like this:

image

Resolves #871 

## TODO

- [ ] #896
- [ ] #897
- [ ] #898

### DIFF
--- a/config/config.ts
+++ b/config/config.ts
@@ -72,6 +72,10 @@ import {
   getFastqUnarchivingManagerStackProps,
   getFastqUnarchivingManagerTableStackProps,
 } from './stacks/fastqUnarchivingManager';
+import {
+  getFastqSyncManagerStackProps,
+  getFastqSyncManagerTableStackProps,
+} from './stacks/fastqSyncManager';
 
 interface EnvironmentConfig {
   name: string;
@@ -115,6 +119,7 @@ export const getEnvironmentConfig = (stage: AppStage): EnvironmentConfig | null 
       accessKeySecretStackProps: getAccessKeySecretStackProps(stage),
       fastqManagerTableStackProps: getFastqManagerTableStackProps(stage),
       fastqUnarchivingManagerTableStackProps: getFastqUnarchivingManagerTableStackProps(),
+      fastqSyncManagerTableStackProps: getFastqSyncManagerTableStackProps(),
     },
     statelessConfig: {
       metadataManagerStackProps: getMetadataManagerStackProps(stage),
@@ -147,6 +152,7 @@ export const getEnvironmentConfig = (stage: AppStage): EnvironmentConfig | null 
       pgDDProps: getPgDDProps(stage),
       fastqManagerStackProps: getFastqManagerStackProps(stage),
       fastqUnarchivingManagerStackProps: getFastqUnarchivingManagerStackProps(stage),
+      fastqSyncManagerStackProps: getFastqSyncManagerStackProps(stage),
     },
   };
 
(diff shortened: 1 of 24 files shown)


alexiswl:
Also ready for review - check out the step functions in development (starting with fastq-sync)

And the Event Bridge rules in development (starting with fastq-sync)

victorskl (approved):

PR #920, opened 2025-03-20
feat(filemanager): ignore directories

### LINKED ISSUE #856: filemanager: handle "directory" objects
Those are objects in S3 that generally don't have any content and are normally only useful for "cosmetic" purposes, to support the idea of "folders". They are created by using an object key ending in `/`.

As they don't represent real "files", but simply simulate folders for end user convenience, they should probably be filtered out / ignored by the FileManager.

Note: 
If such a filter is implemented it's important to only filter/ignore those that have no content as naming objects (with content) with a trailing `/` is a valid operation not limited to the use of simulating folders.

mmalenic:
Closes #856

### Changes
* Ignore directories by filtering out objects that end with "/" and have a size of 0.
    * This is implemented as a env flag for the filemanager ingester function.

I considered handling this inside the EventBridge rule pattern instead, but I think it's not possible. It's not possible to combine ignoring cache objects with `byob-icav2/*/cache/*` and objects that end with `/` with a 0 size because the pattern syntax does not support complex `and`/`or` operations on the same key.

### DIFF
--- a/config/stacks/shared.ts
+++ b/config/stacks/shared.ts
@@ -91,7 +91,20 @@ const getComputeConstructProps = (): ComputeProps => {
   };
 };
 
-const getEventSourceConstructProps = (stage: AppStage): EventSourceProps => {
+export const eventSourcePattern = () => {
+  return {
+    $or: [
+      {
+        size: [{ numeric: ['>', 0] }],
+      },
+      {
+        key: [{ 'anything-but': { suffix: '/' } }],
+      },
+    ],
+  };
+};
+
+export const getEventSourceConstructProps = (stage: AppStage): EventSourceProps => {
   const eventTypes = [
     'Object Created',
     'Object Deleted',
@@ -108,11 +121,14 @@ const getEventSourceConstructProps = (stage: AppStage): EventSourceProps => {
       {
         bucket: oncoanalyserBucket[stage],
         eventTypes,
+        patterns: eventSourcePattern(),
       },
       {
         bucket: icav2PipelineCacheBucket[stage],
         eventTypes,
-        key: [{ 'anything-but': { wildcard: 'byob-icav2/*/cache/*' } }],
+        patterns: {
+          key: [{ 'anything-but': { wildcard: 'byob-icav2/*/cache/*' } }],
+        },
       },
     ],
   };
@@ -121,10 +137,12 @@ const getEventSourceConstructProps = (stage: AppStage): EventSourceProps => {
     props.rules.push({
       bucket: icav2ArchiveAnalysisBucket[stage],
       eventTypes,
+      patterns: eventSourcePattern(),
     });
     props.rules.push({
       bucket: icav2ArchiveFastqBucket[stage],
       eventTypes,
+      patterns: eventSourcePattern(),
     });
   }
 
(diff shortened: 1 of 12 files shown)


victorskl (approved):

reisingerf:
Do we need/want the extra 0 byte check for folders? 
Otherwise it should be simple case to exclude everything that ends with a slash. 
Agreed, then there is a slight chance that an object exists with a name ending in slash and that actually has content. But is that a use case we want or need to support?

mmalenic:
> Agreed, then there is a slight chance that an object exists with a name ending in slash and that actually has content. But is that a use case we want or need to support?

I think we don't have to. But I still think it might not be possible to express this, including both cache buckets and "/" suffixes.

This should work:
```json
{  
  "anything-but": {
    "$or": [
      {
        "key": [{ "suffix": "/" }],
      },
      {
        "key": [{ "wildcard": "byob-icav2/*/cache/*" }],
      },
    ],
  }
}
```
E.g. `NOT (key ends with "/" OR key is cache)` should simplify to `NOT key ends with "/" AND NOT key is cache`.

And AWS does allow it as a valid event pattern, but for some reason it doesn't work when testing it. Not sure if this is a limitation of the event patterns or a bug.

When you remove the `'anything-but'` it works as expected (i.e. it allows all events that end with "/" or are in the cache but).

mmalenic:
There's an open issue to support $and in the AWS event ruler library, which I assume is what AWS uses on EventBridge: https://github.com/aws/event-ruler/issues/69

reisingerf:
Did you try something like this:

```json
"key": [{ "anything-but": { "wildcard": ["byob-icav2/*/cache/*", "*/"] } }]
```

mmalenic:
That works, thanks!

After playing around with it a bit more and avoiding `suffix`, I got it to work with the size requirement too:

I.e. `(key NOT cache AND size > 0) OR (key NOT cache AND key NOT directory)`

```json
"$or": [
  {
    "key": [{ "anything-but": { "wildcard": ["byob-icav2/*/cache/*"] } }],
    "size": [{ "numeric": [">", 0] }]
  },
  {
    "key": [{ "anything-but": { "wildcard": ["byob-icav2/*/cache/*", "*/"] } }]
  },
]
```

mmalenic:
I've updated the PR to use the event pattern instead of doing this inside the filemanager code.

Note that this introduces a dependency on "aws-sdk" for tests. AWS credentials are now needed to successfully run the new test because of [`testEventPattern`](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/eventbridge/command/TestEventPatternCommand/).


I'm also happy to omit the size check if we want to use the simplified code for better readability:

```json
"key": [{ "anything-but": { "wildcard": ["byob-icav2/*/cache/*", "*/"] } }]
```

mmalenic:
Tests will fail until I manually update the pipeline CodeBuild with the `TestEventPattern` action. I'll wait for approval before I do that.

reisingerf (approved):

reisingerf:
Sweet! Thanks!
I think we can leave the size check in the rule. It's a good example we might be able to use elsewhere.

PR #927, opened 2025-03-26
upgrade orcabus to postgres 16

raylrui:
As we discussed in security patrol, our orcabus version is bit lag behind.
"_Your database resources aren't running the latest minor DB engine version. The latest minor version contains the latest security fixes and other improvements._"

so it is time to upgrade postgres to latest version (include testing postgres images).

After testing in DEV, we can upgrade from 15.4 to 16.6 for orcabus databse engine. And it works fine for all ui api call.
As logs show, there will be around 11 mins downtime for this upgrade. ([log link](https://843407916570-ykjyha5d.ap-southeast-2.console.aws.amazon.com/rds/home?region=ap-southeast-2#database:id=orcabus-db;is-cluster=true;tab=logs-and-events))
image

But more testing still need to be done.

### DIFF
--- a/config/stacks/shared.ts
+++ b/config/stacks/shared.ts
@@ -176,8 +176,8 @@ const getDatabaseConstructProps = (stage: AppStage): ConfigurableDatabaseProps =
   const baseConfig = {
     clusterIdentifier: dbClusterIdentifier,
     defaultDatabaseName: 'orcabus',
-    version: AuroraPostgresEngineVersion.VER_15_4,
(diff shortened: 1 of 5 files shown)


raylrui:
After testing, we need add `enableDataApi: true` to enable the default query edit.

Now the RDS in dev account working as expected.

The RDS upgrade will be released with next proper release cycle.

victorskl (approved):
LGTM

reisingerf (approved):

PR #928, opened 2025-03-27
fix: filemanager secret cache

### LINKED ISSUE #925: fix(filemanager): cache secret value for presign user
The filemanager API should cache the secret value when fetching the presigning user credentials, rather than fetching the secret for each presigned URL.

mmalenic:
Closes #925 

### Changes
* Adds caching layer to all `get_secret_value` calls that persists within a Lambda execution environment.


reisingerf:
Sorry, not a Rust guy... perhaps @brainstorm should have a look a this?

Also, how does this compare to solutions documented by AWS like:
https://github.com/aws/aws-secretsmanager-agent/tree/main/aws_secretsmanager_caching

mmalenic:
> Also, how does this compare to solutions documented by AWS like:
https://github.com/aws/aws-secretsmanager-agent/tree/main/aws_secretsmanager_caching

That one is using `RwLock` underneath rather than `Mutex` which I think is the better option. Otherwise behaviour-wise I think it should be fairly similar as all the time-out and max cache entries logic wouldn't be used. I.e. there is one cache entry and it should live for the whole length of the Lambda execution.

It's probably worth switching to that crate anyway as it already implements the locking.

lib/workload/stateless/stacks/filemanager/filemanager-api-lambda/src/main.rs
```diff
@@ -28,6 +29,7 @@ async fn main() -> Result<(), Error> {
         Arc::new(config),
         Arc::new(s3::Client::with_defaults().await),
         Arc::new(sqs::Client::with_defaults().await),
+        Arc::new(Mutex::new(secrets_manager::Client::with_defaults().await)),
```
reisingerf, line 32: Is this where the secret is fetched or the client is initialised?
This code is inside the `main` function. Does that mean it runs inside the handler? 

If so, would simply putting this outside the handler function be easier / better? 

>Take advantage of execution environment reuse to improve the performance of your function. Initialize SDK clients and database connections outside of the function handler...

(from: https://docs.aws.amazon.com/lambda/latest/dg/rust-handler.html#rust-best-practices)

mmalenic: Yes, that's where the client gets initialized. This is put into the `AppState` struct which is shared for all requests. 

> This code is inside the main function. Does that mean it runs inside the handler?

Everything inside the `main` function before [`run(app).await`](https://github.com/umccr/orcabus/blob/0423e255306fc71929d30ee1c31de6aeb5d8aeec/lib/workload/stateless/stacks/filemanager/filemanager-api-lambda/src/main.rs#L40) should only be initialized once when the Lambda execution loads. [`app`](https://github.com/umccr/orcabus/blob/0423e255306fc71929d30ee1c31de6aeb5d8aeec/lib/workload/stateless/stacks/filemanager/filemanager-api-lambda/src/main.rs#L37-L38) in this context is the handler which is what axum uses to process requests. This state is cloned for each request, but that's okay because it still uses the same cache variable which is stored on heap inside a reference counting pointer - [`Arc`](https://doc.rust-lang.org/std/sync/struct.Arc.html).

The [`Mutex`](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html) locks the client exclusively so that it's mutable. This is required to modify the state as it has to [insert](https://github.com/umccr/orcabus/blob/0423e255306fc71929d30ee1c31de6aeb5d8aeec/lib/workload/stateless/stacks/filemanager/filemanager/src/clients/aws/secrets_manager.rs#L47) an entry when fetching a secret (or get an existing entry).

The `aws_secretsmanager_caching` crate does a similar thing except the lock is inside that crate and it's a [`RwLock`](https://github.com/aws/aws-secretsmanager-agent/blob/43d15f2b31de07056cff7dc5613ec75173ab2ad1/aws_secretsmanager_caching/src/lib.rs#L35C12-L35C18). I think this should be better as it doesn't lock exclusively for reads - allowing multiple at the same time.

As an aside, this pattern (i.e. `Arc<Mutex<T>>` or `Arc<RwLock<T>>`) allows cloning/modifying shared state across threads from a regular `&` reference, rather than requiring `&mut`. E.g. noting the function that the `aws_secretsmanager_caching` makes public has a [`&self`](https://github.com/aws/aws-secretsmanager-agent/blob/43d15f2b31de07056cff7dc5613ec75173ab2ad1/aws_secretsmanager_caching/src/lib.rs#L160) rather than `&mut self`, even though that function does modify the state of self (through the RwLock). This normally wouldn't be allowed in Rust. I think `Mutex`/`RwLock` uses some [unsafe](https://doc.rust-lang.org/src/std/sync/poison/rwlock.rs.html#461-464) code underneath to convince the compiler that this is okay.

Hopefully that makes sense, let me know!

victorskl (approved):

brainstorm (approved):

reisingerf (approved):

PR #937, opened 2025-03-31
Fix: Failing dependency check

williamputraintan:
Upgrade the relevant dependency


package.json
```diff
@@ -27,7 +27,7 @@
     "@aws-cdk/aws-lambda-python-alpha": "2.177.0-alpha.0",
     "@aws-cdk/aws-pipes-alpha": "2.177.0-alpha.0",
     "@aws-cdk/aws-pipes-sources-alpha": "2.177.0-alpha.0",
-    "aws-cdk-lib": "2.177.0",
+    "aws-cdk-lib": "2.186.0",
```
victorskl, line 30: If this is just for cdk CLI hotfix purpose and, have not majorly broken yet then pls go for it, Will. 

Typically, I'd advice/wish to maintain (and update at the same time) all `@aws-cdk/*` as well in the same major/minor/patch version. The one those up above.

victorskl (approved):
LGTM with minor note. Trust you!

PR #941, opened 2025-04-01
Enable corepack before all yarn --install commands

alexiswl:
By removing the .yarn/yarn--version.cjs file I think corepack enable is needed order for the build commands to work - see 
https://umccr.slack.com/archives/C070B52CHT9/p1743493591199569?thread_ts=1743387409.116639&cid=C070B52CHT9

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -7,6 +7,7 @@ tasks:
     init: |
       pre-commit install
       pre-commit run
+      corepack enable
       yarn install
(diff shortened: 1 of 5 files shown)


williamputraintan:
@victorskl I guess we should now abandon the `.yarn/yarn--version.cjs` approach

victorskl:
Should be fine as we are splitting the repo soon. Just watch out on "it works on my local" syndrome at CI environment (if any)...

PR #943, opened 2025-04-01
Fix - (SRM) ignore status changed after sequece run succeed

raylrui:
modify logic rules for status_changed,
status_changed to be true if:
    - new sequence is created
    - sequence status has changed and current sequence status is not SequenceStatus.SUCCEEDED (SequenceStatus.SUCCEEDED means sequence run is succeeded, any failed or aborted state after that is coming from analysis stage which is not relevant to this sequence run)


raylrui:
After this rule approved, I will modify the PROD latest records 'false' failed sequence run.

reisingerf (approved):

raylrui:
PROD record changed for record `instrument_run_id='250328_A01052_0258_AHFGM7DSXF'`

PR #955, opened 2025-04-11
fix: filemanager sequencer ordering

### LINKED ISSUE #949: filemanager: crawl synthetic events are treated as duplicates
The filemanager relies on the "sequencer" values from S3 events to de-duplicate events. This works well for regular events that AWS generates, but it fails for synthetically generated events from crawls, as the sequencer is [always](https://github.com/umccr/orcabus/blob/f0fe4ce1c872feea99270dbdb0ab1924d007ff56/lib/workload/stateless/stacks/filemanager/filemanager/src/events/aws/crawl.rs#L77) set to the same value. This means that repeated crawls on the same object will always be considered duplicates and not ingested, even if the storage class or other properties change. 

Noting that this only affects crawls. Regular events, such as `s3:LifecycleTransition` are correctly handled.

### LINKED ISSUE #950: ingestId changes for file when storage class changes when moving from standard to intelligent tiering
No urgency on this one but will be something that we will need to discuss next week as I think this puts in a few roadblocks.

Consider the following file that is pulled out of archive (with the ingest id 019387f1-7455-74a2-a763-e627b5fbce06), which is what the fastq manager points to.

```
s3://archive-prod-fastq-503977275616-ap-southeast-2/v1/year=2024/month=01/240105_A00130_0284_AHWLVYDSX7/20241122d876a770/WGS_TsqNano/PRJ231302_L2301519_S4_L001_R1_001.fastq.ora
```

To

```
s3://pipeline-prod-cache-503977275616-ap-southeast-2/byob-icav2/production/restored/14d/year=2025/month=04/day=03/728c6b30-df79-47b5-9b70-59c098a11455/240105_A00130_0284_AHWLVYDSX7/PRJ231302_L2301519_S4_L001_R1_001.fastq.ora
```

The ingest id is maintained as `019387f1-7455-74a2-a763-e627b5fbce06`

However, a day later the Object's storage class moves from Standard to 'IntelligentTiering' and as a result, the filemanager generates a new event with a new ingest id for this file and the original ingest id is no longer the 'current' ingest id for that file. 

I also don't think that the currentState boolean is set correctly. 

```
curl --fail-with-body --silent --show-error --location \
  --request 'GET' \
  --header 'Accept: application/json' \
  --header "Authorization: Bearer $ORCABUS_TOKEN" \
  --url 'https://file.prod.umccr.org/api/v1/s3?page=1&rowsPerPage=1000&caseSensitive=true&currentState=false&eventType=Created&bucket=pipeline-prod-cache-503977275616-ap-southeast-2&key=byob-icav2%2Fproduction%2Frestored%2F14d%2Fyear%3D2025%2Fmonth%3D04%2Fday%3D03%2F728c6b30-df79-47b5-9b70-59c098a11455%2F240105_A00130_0284_AHWLVYDSX7%2FPRJ231302_L2301519_S4_L001_R1_001.fastq.ora'
```

Gives

```json
{
  "links": {
    "previous": null,
    "next": null
  },
  "pagination": {
    "count": 2,
    "page": 1,
    "rowsPerPage": 1000
  },
  "results": [
    {
      "s3ObjectId": "0195fb8c-00c8-7c31-bf30-e99f6e9db01a",
      "eventType": "Created",
      "bucket": "pipeline-prod-cache-503977275616-ap-southeast-2",
      "key": "byob-icav2/production/restored/14d/year=2025/month=04/day=03/728c6b30-df79-47b5-9b70-59c098a11455/240105_A00130_0284_AHWLVYDSX7/PRJ231302_L2301519_S4_L001_R1_001.fastq.ora",
      "versionId": "gjpHd6x5tjmu.HA_8j5tZy6KyoGsK4OO",
      "eventTime": "2025-04-03T12:06:50Z",
      "size": 8252959343,
      "sha256": null,
      "lastModifiedDate": "2025-04-03T12:03:59Z",
      "eTag": "\"809440dafb05109a7617e1fc4611a98b-2\"",
      "storageClass": "Standard",
      "sequencer": "0067EE792E423105A3",
      "isDeleteMarker": false,
      "numberDuplicateEvents": 0,
      "attributes": null,
      "deletedDate": null,
      "deletedSequencer": null,
      "numberReordered": 0,
      "ingestId": "019387f1-7455-74a2-a763-e627b5fbce06",
      "isCurrentState": false,
      "reason": "CreatedCompleteMultipartUpload",
      "archiveStatus": null,
      "isAccessible": false
    },
    {
      "s3ObjectId": "0196002a-f245-7af1-b3e4-4f1eef6379bd",
      "eventType": "Created",
      "bucket": "pipeline-prod-cache-503977275616-ap-southeast-2",
      "key": "byob-icav2/production/restored/14d/year=2025/month=04/day=03/728c6b30-df79-47b5-9b70-59c098a11455/240105_A00130_0284_AHWLVYDSX7/PRJ231302_L2301519_S4_L001_R1_001.fastq.ora",
      "versionId": "gjpHd6x5tjmu.HA_8j5tZy6KyoGsK4OO",
      "eventTime": "2025-04-04T09:38:55Z",
      "size": 8252959343,
      "sha256": null,
      "lastModifiedDate": "2025-04-03T12:03:59Z",
      "eTag": "\"809440dafb05109a7617e1fc4611a98b-2\"",
      "storageClass": "IntelligentTiering",
      "sequencer": null,
      "isDeleteMarker": false,
      "numberDuplicateEvents": 0,
      "attributes": null,
      "deletedDate": null,
      "deletedSequencer": null,
      "numberReordered": 0,
      "ingestId": "0195fb8c-0117-7800-af9d-c0ef8d3cb395",
      "isCurrentState": true,
      "reason": "StorageClassChanged",
      "archiveStatus": null,
      "isAccessible": true
    }
  ]
}
```

But this also returns the 'isCurrentState=true' despite currentState being set to false in the params. 

I'm not sure what to do here, this is causing some problems for the data-sharing manager since it relies on the fastq manager to determine what files are in archive and which ones are not. The fastq manager uses the ingest ids to determine the s3 uris. But because the ingest id of the extracted data is no longer the current state, the fastq manager points back to the files in archive.  So unless we send out the data before data is converted to intelligent tiering state, the fastq manager points back to the original files in the archive, making the user scratch their head 'I swear I unarchived those already'.  And duplicating a lot of data in the restore directory. 

Alternatively, we could resolve the ingest ids with a bit more intelligence where currentState is set to false, 

* find all uris for a given ingest id,
* re-query the file manager by the uris
* for each uri, return only states since the most recent delete marker.
* Then sort each uri by the storage class
* This seems like significantly more work in the logic and probably a lot more strain on the file manager API since it would require a 'double' query to return the responses for the uris.

This will also be a problem when we start automating the fastq manager to ingest outputs from the BSSHFastqCopy complete service since this will happen while that data is still in the standard class?

Or maybe we make an exception of the transition from standard to intelligent tiering the ingest id is kept the same?

mmalenic:
Closes #950
Closes #949

### Context

Some AWS events do not contain sequencer values, like lifecycle transition events. Also, crawls on objects do not have sequencer values as they represent point-in-time generated events. This is an issue for the filemanager because it uses the `sequencer` to order events and update attributes like `is_current_state`. If a null sequencer is inserted into the database it is always ordered after any non-null events, which means that when a newer event comes in, it does not override the existing record's `is_current_state`. 

This was the cause of a few bugs with both the crawl functionality and storage class events from lifecycle transitions.

### Changes

Now, a sequencer value is always present. If an event doesn't have one, or if it is a synthetic event from a crawl, a new sequencer value is generated. The sequencer is generated so that events are always compared by sequencer first, and then by event time.

#### What this means
* AWS-native sequencers are preferred for ordering if they are available. This matches the existing logic.
* If there is no sequencer, the inserted event overrides the current state of records, but any new events (either with AWS-native sequencers or generated sequencers) override that event.

#### The way this is implemented
* Sequencers are not modified if they are present with the incoming event.
* If no sequencer is present a new one is generated by:
    * Taking an existing sequencer from the "current" state of the database records. Padding it with a lot of zeros, and then appending a `u64` at the end. 
    * If a "current" sequencer doesn't exist (because this is a new key/bucket combination), the smallest possible all "0" sequencer is used.
* When a new event comes in, if it has an AWS-native sequencer, it will always be greater than any generated sequencer before it. If it's another synthetic event (or lifecycle transition), then the `u64` at the end of the padded sequencer is incremented. 
* The padding is necessary because AWS says that sequencers can have different lengths. As long as the padding is large enough that a sequencer value from AWS will never be bigger, everything works.
    * There are only sequencers that are 18 characters long in the current dev filemanager database.
    
#### Some other notes
* The code path that executes this is only triggered if the sequencer is null. This means that the ingester stays the same for all events except: `s3:ObjectRestore:*`, `s3:LifecycleTransition` and crawls.

### Other related fixes
* The version_id is correctly set when fetching data from the SDK. This was causing issues between non-versioned and versioned buckets, because a version_id of "null" is used for non-versioned buckets.
* I've reverted the changes from https://github.com/umccr/orcabus/pull/930. The filemanager API needs the same permissions as the ingest role because the crawl function performs similar actions as the ingester.
* `FILEMANAGER_INGESTER_TAG_NAME` is set for the API function.
* Generated sequencers are not supported with `PAIRED_INGEST_MODE` and I've disallowed this in the code. This mode is not used being used anyway.


reisingerf (approved):

victorskl (approved):
LGTM.

Thanks for documenting. 👍

brainstorm:
> Some AWS events do not contain sequencer values, like lifecycle transition events.

Fascinating! Are those lifecycle transition events the only ones missing sequencer values?

brainstorm (approved):

alexiswl (approved):

mmalenic:
> Fascinating! Are those lifecycle transition events the only ones missing sequencer values?

Yes! Those ones, i.e. "s3:LifecycleTransition", "s3:ObjectRestore*" and "s3:IntelligentTiering" don't have sequencer values.

Interestingly, "s3:LifecycleExpiration" does have a sequencer value though. 

I'm not entirely sure why. I guess AWS doesn't think the ordering of the lifecycle events is that important?

mmalenic:
Some of the other events might also not have sequencers, e.g. "ObjectTagging:*", but filemanager isn't ingesting those for now anyway.

PR #958, opened 2025-04-14
Fix #957

### LINKED ISSUE #957: BclConvertManager - race condition when assigning portal_run_id
See 
https://orcaui.umccr.org/runs/workflow/wfr.01JRK6NA2GC0T5SS10MQ1JV5EH
and
https://orcaui.umccr.org/runs/workflow/wfr.01JRK6NA2GAT2P30ZJS1W91SE0

result in two different portal run ids (`202504112d1ceb00`, `20250411bf0ca669`) event though they relate to the same BCL Convert run (`Tsqn250407_10Apr25_46522d_736b75 `).

This is due to a race condition when incoming ICA events come at the same time.
As lambdas now run in parallel it's easiest to avoid the race condition by removing the parallel lambda execution. (e.g. set reserved concurrency to 1)

reisingerf:
Fixes: https://github.com/umccr/orcabus/issues/957

### DIFF
--- a/lib/workload/stateless/stacks/bclconvert-manager/deploy/stack.ts
+++ b/lib/workload/stateless/stacks/bclconvert-manager/deploy/stack.ts
@@ -104,6 +104,7 @@ export class BclConvertManagerStack extends Stack {
         ICAV2_ACCESS_TOKEN_SECRET_ID: icav2JwtSecretsManagerObj.secretName,
       },
       architecture: Architecture.ARM_64,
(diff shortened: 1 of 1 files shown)


raylrui:
yup, race condition could happened when incoming ICA events come at the same time. good learn of reserving concurrency.

raylrui (approved):
LGTM. maybe add a comma `,` at the end of the attribute can remove the prettier error from current rules, like `reservedConcurrentExecutions: 1,`

PR #966, opened 2025-04-17
Fix ora inputs sfn for cttso workflow input handler

alexiswl:
Small error when trying to run down decompressing oras path


victorskl (commented):
LGTM

PR #997, opened 2025-05-19
refactor(stateless): detach direct dependency on the filemanager stack

mmalenic:
### Changes
* Removes direct stack-based dependency on filemanager from htsget and fmannotator.
* Update htsget-rs construct dependency to support `IRole` instead of `Role`.

I need to remove this because otherwise when migrating filemanager to the new [repo](https://github.com/OrcaBus/service-filemanager), I can't update the new stack because CloudFormation expects that the FileManager stack exports variables for the domain name and ingest role. This way, the dependency is more loosely coupled (with constants/parameters) so it won't refuse to update in CloudFormation.

This shouldn't result in any changes to the behaviour of the services.

### DIFF
--- a/lib/workload/stateless/statelessStackCollectionClass.ts
+++ b/lib/workload/stateless/statelessStackCollectionClass.ts
@@ -48,7 +48,7 @@ import {
   OraCompressionIcav2PipelineManagerStackProps,
 } from './stacks/ora-compression-manager/deploy';
 
-import { FMAnnotator, FMAnnotatorConfigurableProps } from './stacks/fmannotator/deploy/stack';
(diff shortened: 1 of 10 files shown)


lib/workload/stateless/statelessStackCollectionClass.ts
```diff
@@ -333,7 +331,6 @@ export class StatelessStackCollection {
     this.htsgetStack = new HtsgetStack(scope, 'HtsgetStack', {
       ...this.createTemplateProps(env, 'HtsgetStack'),
       ...statelessConfiguration.htsgetProps,
-      role: fileManagerStack.ingestRole,
```
mmalenic, line 336: ... and here.

victorskl:
Reviewing...

victorskl (approved):
LGTM

PR #1023, opened 2025-06-16
Added ora reference support to WGTS QC pipeline

alexiswl:
Bit of a hacky PR (since we're not bringing wgts-qc forward), to add in the ora reference tarball into the icav2 workflow inputs.  

Only an issue now since BCLConvert now outputs ORA files


victorskl (approved):
LGTM

PR #1024, opened 2025-06-16
Add ecs managed policy for fargate tasks

alexiswl:
Allows AWS Guard Duty agent to run in parallel,

Also small fix to function name in fastq manager.
And remove fastqSet/ prefix duplicate from validate endpoints


victorskl (approved):
LGTM

